PR #17, opened 2023-09-26
Read token secret from CLICKHOUSE_TOKEN_SECRET environment variable

swgillespie:
The code as written reads a default value for tokenSecret by reading the 'CLICKHOUSE_TOKEN_KEY' environment variable, same as the token key; I think this is a typo and this commit reads the secret from a separate environment variable.


Kinzeng (approved):
good catch, thanks for the PR!

PR #35, opened 2023-12-21
implement PrivateLink flow

Kinzeng:
this PR implements the complete the entire PrivateLink flow using a new resource `clickhouse_private_endpoint_registration`. users can add the PrivateLink endpoint ids into this resource and reference the `private_endpoint_ids` attribute in any of their services.


clickhouse/client.go
```diff
@@ -183,12 +204,31 @@ func (c *Client) GetService(serviceId string) (*Service, error) {
 		return nil, err
 	}
 
-	return &serviceResponse.Result, nil
+	service := serviceResponse.Result
+
+	req, err = http.NewRequest("GET", c.getServicePath(serviceId, "/privateEndpointConfig"), nil)
```
tsolodov, line 209: proposed new schema [here](https://github.com/ClickHouse/control-plane/issues/6235)

we should use new regional endpoint to retrieve this data.

examples/privatelink/main.tf
```diff
@@ -0,0 +1,68 @@
...
+
+  min_total_memory_gb  = 24
+  max_total_memory_gb  = 360
+  idle_timeout_minutes = 5
+
+  # use registered endpoint ids
+  private_endpoint_ids = clickhouse_private_endpoint_registration.private_endpoints.private_endpoint_ids
+}
+
+output "service_endpoints" {
```
tsolodov, line 62: we should probably add TF code examples for PL config(commented code?):

something like:
```
resource "aws_vpc_endpoint" "this" {
  vpc_id            = var.vpc_id
  service_name      = clickhouse_service.service.endpoints <- it should use regional property, not instance's property to pass this info.
  vpc_endpoint_type = "Interface"
  security_group_ids = [
    Var.security_group_id1,var.security_group_id2, var.security_group_id3,
  ]
  subnet_ids          = [var.subnet_id1,var.subnet_id2,var.subnet_id3]
  private_dns_enabled = false
}
```

examples/privatelink/main.tf
```diff
@@ -0,0 +1,68 @@
...
+      description = "Test IP"
+    }
+  ]
+
+  min_total_memory_gb  = 24
+  max_total_memory_gb  = 360
+  idle_timeout_minutes = 5
+
+  # use registered endpoint ids
+  private_endpoint_ids = clickhouse_private_endpoint_registration.private_endpoints.private_endpoint_ids
```
tsolodov, line 59: do you want to use a bit different format? `clickhouse_private_endpoint_registration.private_endpoints.aws.private_endpoint_ids` ?

What if customer defined this:
```
resource "clickhouse_private_endpoint_registration" "private_endpoints" {
  private_endpoints = [
    {
      cloud_provider = "aws"
      id             = "vpce-abcdef12345678910"
      region         = "us-east-1"
    },
    {
      cloud_provider = "gcp"
      id             = "1233"
      region         = "us-east1"
    }
  ]
}
```

tsolodov: plz ignore me, it can be fixed by adding another `clickhouse_private_endpoint_registration` resource

examples/privatelink/main.tf
tstraley: I'd want to see this in the example.

This example shows a current issue with this implementation - there is a dependency loop where this `vpc_endpoint` requires the `clickhouse_service` to be created already (maybe that is what the comment about using regional property is about?) but the `clickhouse_service` also depends on the generated ID of this endpoint being created first.

guykoh:
@Kinzeng what's the status of this PR? When do you expect to merge this?

examples/privatelink/main.tf
tsolodov: Added

tsolodov (approved):
I tested PrivateLink / PSC creation using 0.0.5-beta.7 version. Everything works as expected.

Added example TF code that configures AWS/GCP infra & ClickHouse cloud.

Issue #30, opened 2023-12-13
Issues supporting AWS Privatelink / GCP Private Svc Connect configurations


tstraley:
Hi - I would like to use terraform to create a Production tier clickhouse service that is connected to my AWS VPC via privatelink, per the documentation here: https://clickhouse.com/docs/en/manage/security/aws-privatelink

I've faced numerous issues in accomplishing this, and want to share my findings for how this provider can be fixed:

## 1. `ip_access` cannot be empty array

With privatelink, we don't want any public access - only private access - so we don't want to add any IPs to this public access list. When setting `ip_access = []` the apply fails with
```
╷
│ Error: Error creating service
│ 
│   with module.clickhouse.clickhouse_service.main,
│   on ../libs/clickhouse/main.tf line 8, in resource "clickhouse_service" "main":
│    8: resource "clickhouse_service" "main" {
│ 
│ Could not create service, unexpected error: status: 400, body:
│ {"requestId":"b6e2d41a-0c59-4943-bdd9-f89511fbb38f","error":"BAD_REQUEST:
│ request body.ipAccessList value is not an array: undefined","status":400}
╵
```

As a workaround, you can assign an ip_acess source during create, and then later remove it (but then you need to deal with this issue https://github.com/ClickHouse/terraform-provider-clickhouse/issues/29)

Alternative workaround is to use something like `ip_access = [{ source = "127.0.0.1", description = "No external public access" }]` but this is obviously not ideal.

## 2. The `privateEndpointConfig` info about the service is not available in attributes

To set up privatelink, we need the `endpointServiceId` which is from the `v1/organizations/$ORG_ID/services/$INSTANCE_ID/privateEndpointConfig` API. This seems like something that could be included in the `clickhouse_service` resource read-only attributes, but it isn't.

As a workaround, can do:
```
data "http" "private_endpoint_config" {
  url = format(
    "https://api.clickhouse.cloud/v1/organizations/%s/services/%s/privateEndpointConfig",
    var.clickhouse_cloud_org_id,
    clickhouse_service.main.id
  )

  request_headers = { Authorization = local.basic_auth }
}
```

But this is obviously not ideal.

## 3. There is no way to provide the VPC endpoint ID to the org or service configuration

To establish trust and allow access from my AWS VPC, we need to add our VPC endpoint ID created for the private link to the Org config and the service.

[...]

PR #74, opened 2024-07-29
Issues supporting aws privatelink

egor-krv:
This PR solves this [issue](https://github.com/ClickHouse/terraform-provider-clickhouse/issues/30). To solve the issue we need to have a ClickHouse service provisioned first before attaching the endpoints because the endpoints creation depends on the ClickHouse service itself. I moved endpoint attachment to a new resource so it happens after the endpoints are created.

### DIFF
--- a/examples/PrivateLink/aws.tf
+++ b/examples/PrivateLink/aws.tf
@@ -1,19 +1,9 @@
-variable "aws_key" {
-  type = string
-}
-
-variable "aws_secret" {
-  type = string
-}
-
 variable "aws_region" {
   type = string
 }
 
 provider "aws" {
   region     = var.aws_region
-  access_key = var.aws_key
-  secret_key = var.aws_secret
 }
 
 variable "vpc_foo_id" {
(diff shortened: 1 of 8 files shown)


examples/PrivateLink/main.tf
```diff
@@ -1,7 +1,7 @@
 terraform {
   required_providers {
     clickhouse = {
-      version = "0.0.5"
+      version = "0.0.10"
```
whites11, line 4: technically this example won't work until this PR is merged, and thus released with a future version of the provider, right?

examples/PrivateLink/aws.tf
```diff
@@ -1,19 +1,9 @@
...
-  type = string
-}
-
 variable "aws_region" {
   type = string
 }
 
 provider "aws" {
   region     = var.aws_region
-  access_key = var.aws_key
```
whites11, line 15: How is this change related with this PR?

examples/PrivateLink/main.tf
egor-krv: This is the current version so we are all good here.

examples/PrivateLink/aws.tf
egor-krv: Just modifying the example so it works from environment vars so it is easier to run

whites11 (approved):

PR #89, opened 2024-07-31
add "import" section to docs/resources/service.md

daniel-habib:


docs/resources/service.md
```diff
@@ -74,3 +74,11 @@ Read-Only:
 
 - `endpoint_service_id` (String) Unique identifier of the interface endpoint you created in your VPC with the AWS(Service Name) or GCP(Target Service) resource
 - `private_dns_hostname` (String) Private DNS Hostname of the VPC you created
+
+## Import
```
nikita-vanyasin, line 78: Hi @daniel-habib 
Thanks for your contribution. 
Please note this file is auto generated.

To add new section about importing state, please add it to index.md

nikita-vanyasin (changes requested):

docs/resources/service.md
daniel-habib: auto-generate: cool
i see it is done using `tfplugindocs generate --provider-name=clickhouse` 

i have not worked with that before.
per their documentation to achive that i should create some import file and then set `.HasImport` and `.ImportFile`.
i was expecting some templates dir and in it resources.md.tmpl or some other template name.

when you say "add it to index.md" do you mean docs/index.md or some place else? a pointer to an example would suffice.

whites11:
I made the changes to add import example here: https://github.com/ClickHouse/terraform-provider-clickhouse/pull/92

PR #97, opened 2024-08-01
move api client code from clickhouse package to own internal one and …

whites11:
…add interface for mocking purposes

This PR is trying to achieve:

- package isolation by moving CH (control plane) API client to own internal package.
- mockability to allow testing of provider by mocking the CH api client

### DIFF
--- a/clickhouse/service.go
+++ b/clickhouse/service.go
@@ -2,6 +2,10 @@ package clickhouse
 
 import (
 	"context"
+	"crypto/sha1"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/hex"
 	"errors"
 	"strings"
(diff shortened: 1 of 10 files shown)


clickhouse/service.go
```diff
@@ -931,3 +937,15 @@ func (r *ServiceResource) syncServiceState(ctx context.Context, state *ServiceRe
 
 	return nil
 }
+
+func ServicePasswordUpdateFromPlainPassword(password string) api.ServicePasswordUpdate {
```
whites11, line 941: This was in the `client` file but it's a helper function. Leaving it here for now but it'll likely go away eventually

nikita-vanyasin (approved):

PR #100, opened 2024-08-02
base ground for testing

whites11:
…add interface for mocking purposes

more Resource model to own file and implement Equals function to allow comparison between instances.

Implement unit test to check Equals function

### DIFF
--- a/clickhouse/service.go
+++ b/clickhouse/service.go
@@ -8,9 +8,8 @@ import (
 	"encoding/hex"
 	"errors"
 	"strings"
-	"time"
-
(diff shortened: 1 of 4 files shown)


clickhouse/service.go
```diff
@@ -41,30 +41,6 @@ type ServiceResource struct {
 	client api.Client
 }
 
-type ServiceResourceModel struct {
```
whites11, line 44: moved to own file

nikita-vanyasin (approved):

PR #106, opened 2024-08-05
Add unit test to check mapping of API responses to terraform status

whites11:


clickhouse/service.go
```diff
@@ -56,7 +56,8 @@ type IpAccessModel struct {
 }
 
 func (s IpAccessModel) Equal(other IpAccessModel) bool {
-	return false
+	return s.Source.ValueString() == other.Source.ValueString() &&
```
whites11, line 59: forgot to push implementation on a previous PR

PR #111, opened 2024-08-06
Refactoring

whites11:


pkg/resource/models/service_resource.go
```diff
@@ -0,0 +1,125 @@
...
+	IAMRole                         types.String `tfsdk:"iam_role"`
+	LastUpdated                     types.String `tfsdk:"last_updated"`
+	PrivateEndpointConfig           types.Object `tfsdk:"private_endpoint_config"`
+	PrivateEndpointIds              types.List   `tfsdk:"private_endpoint_ids"`
+	EncryptionKey                   types.String `tfsdk:"encryption_key"`
+	EncryptionAssumedRoleIdentifier types.String `tfsdk:"encryption_assumed_role_identifier"`
+}
+
+func (m *ServiceResourceModel) Equals(b ServiceResourceModel) bool {
+	if !m.ID.Equal(b.ID) ||
```
nikita-vanyasin, line 100: An idea for next step: these Equal functions can be generated with something like https://github.com/awalterschulze/goderive/tree/main/example/plugin/equal

pkg/resource/models/service_resource_test.go
```diff
@@ -283,14 +249,14 @@ func getBaseModel() ServiceResourceModel {
 		Region:                          types.StringValue(""),
 		Tier:                            types.StringValue(""),
 		IdleScaling:                     types.Bool{},
-		IpAccessList:                    make([]IPAccessModel, 0),
+		IpAccessList:                    tfutils.CreateEmptyList(IpAccessList{}.ObjectType()),
```
nikita-vanyasin, line 252:
```suggestion
		IPAccessList:                    tfutils.CreateEmptyList(IpAccessList{}.ObjectType()),
```

nikita-vanyasin (approved):

pkg/resource/models/service_resource_test.go
whites11: done thanks

pkg/resource/models/service_resource.go
whites11: nice one

Issue #71, opened 2024-07-02
clickhouse_private_endpoint_registration always update in-place the id in terraform plan/apply


yardenws:
#Title# 

When using the resource "clickhouse_private_endpoint_registartion"
Every time I run terraform plan / apply it says there are changes to the infrastructure. 
happens in both scenarios:
* Getting the ID from aws_vpc_endpoint resource.
* hardcoded the ID. 

**Code example:**
resource "clickhouse_private_endpoint_registration" "private_endpoints" {
    cloud_provider = "aws"
    id             = "vpce-xxxx"
    region         = var.region
}
----------------------
OR 

resource "aws_vpc_endpoint" "legacy2clickhouse" {
  provider            = aws.remote
  vpc_id              = var.legacy_vpc_id
  service_name        = data.clickhouse_private_endpoint_config.service_id.endpoint_service_id
  vpc_endpoint_type   = "Interface"
  security_group_ids  = [for sg in data.aws_security_group.legacy_security_group : sg.id]
  subnet_ids          = [for subnet in data.aws_subnet.legacy_subnet : subnet.id]
  private_dns_enabled = true
  tags = {
    Name = "${var.environment} Clickhouse Legacy Peering"
  }
}

resource "clickhouse_private_endpoint_registration" "private_endpoints" {
    cloud_provider = "aws"
    id             = aws_vpc_endpoint.legacy2clickhouse.id
    region         = var.region
}
----------------------

**Output example:**
Terraform will perform the following actions:

  # module.site.module.clickhouse[0].clickhouse_private_endpoint_registration.private_endpoints will be updated in-place
  ~ resource "clickhouse_private_endpoint_registration" "private_endpoints" {
        id             = "vpce-xxxxxxx"
        # (3 unchanged attributes hidden)
    }

smstone:
Hello @yardenws,

I attempted to recreate the issue you've described above but was not able to. I used the configuration from the examples directory and used the latest ClickHouse provider version `v0.0.10`.

My environment setup is as follows:
```
$ terraform version
Terraform v1.9.3
on darwin_arm64
+ provider registry.terraform.io/clickhouse/clickhouse v0.0.10
+ provider registry.terraform.io/hashicorp/aws v5.60.0
```

Can you please let me know which versions you're using when this happens? If you're not on the latest, can you please try them to see if it addresses this issue?

Thank you,
Steven

yardenws:
[...]

PR #114, opened 2024-08-06
Remove private_endpoint_attachment resource

whites11:
See https://github.com/ClickHouse/terraform-provider-clickhouse/issues/71#issuecomment-2271549956

The private_endpoint_attachment fights with the `private_endpoint_ids` field of the `clickhouse_service` resource. They can never work together so I am going to remove the recently added `private_endpoint_attachment` resource.

I also updated the private link sample to make it work


whites11:
CC @egor-krv as you added this resource recently

Issue #137, opened 2024-08-19
It's not possible to use private endpoint for the first service in an org


whites11:
It's necessary to specify `depends_on` on every resource, but there is a circular dependency:

"clickhouse_private_endpoint_config" data source needs "clickhouse_service" resource
"clickhouse_service" needs "clickhouse_private_endpoint_registration" resource
"clickhouse_private_endpoint_registration" needs "aws_vpc_endpoint" resource
"aws_vpc_endpoint" needs "clickhouse_private_endpoint_config" data source

Without specifying any of the `depends_on` clause, `terraform plan` fails with the following error

```
Planning failed. Terraform encountered an error while generating this plan.

╷
│ Error: failed get
│ 
│   with data.clickhouse_private_endpoint_config.endpoint_config,
│   on main.tf line 54, in data "clickhouse_private_endpoint_config" "endpoint_config":
│   54: data "clickhouse_private_endpoint_config" "endpoint_config" {
│ 
│ error getting privateEndpointConfig: status: 400, body: {"requestId":"9c1f9551-d355-4046-acfc-4ecf5be37ea8","error":"BAD_REQUEST: organization has no created instances in aws eu-west-1","status":400}
```

whites11:
Possible solutions:

1) prevent the API error `BAD_REQUEST: organization has no created instances in aws eu-west-1` when an organization has no services
2) decouple service creation and `private_endpoint_ids` association. Basically revert `https://github.com/ClickHouse/terraform-provider-clickhouse/pull/114` and make it work (requires changes to the backend API).
3) Make the existence of a service a requirement and make `ŧerraform plan` fail in a better way.

whites11:
According to control plane team, number 1 is not possible because:

```
Yeah unfortunately we have this limitation. We can't ask data plane about this information until an instance exists in that region. I believe this has to do with DP's cellular architecture -- the result would depend on the cell assigned to the (organization, region) pair, and that assignment doesn't occur until an instance is created there.
```

whites11:
the only viable solution is number 2.
I am working on that.

unfortunately this will be a breaking change for out customers, but there is no way around this

whites11:
fixed in latest main, with a major release pending

PR #138, opened 2024-08-20
Remove private endpoints from service

whites11:
Towards: https://github.com/ClickHouse/terraform-provider-clickhouse/issues/137

Deprecate and avoid using `private_endpoint_ids` attribute for the `service` resource.

The feature will be replaced by a standalone resource in a followup PR.

Of course this is a breaking change and will be released in a major version.

### DIFF
--- a/pkg/resource/private_endpoint_registration.go
+++ b/pkg/resource/private_endpoint_registration.go
@@ -126,19 +126,15 @@ func (r *PrivateEndpointRegistrationResource) Read(ctx context.Context, req reso
 		}
 	}
 
-	if privateEndpoint == nil {
-		resp.Diagnostics.AddError("Private endpoint not found", "Could not find private endpoint in org registration")
-		return
-	}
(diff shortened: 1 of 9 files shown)


pkg/resource/private_endpoint_registration.go
```diff
@@ -126,19 +126,15 @@ func (r *PrivateEndpointRegistrationResource) Read(ctx context.Context, req reso
 		}
 	}
 
-	if privateEndpoint == nil {
-		resp.Diagnostics.AddError("Private endpoint not found", "Could not find private endpoint in org registration")
```
egor-krv, line 130: Maybe include private endpoint Name here so the user knows which one?

egor-krv (approved):

pkg/resource/private_endpoint_registration.go
whites11: this code was removed.

Issue #133, opened 2024-08-09
Handle 429 and 500 in API client


whites11:
While running e2e tests, if we hit the api too hard, the destroy phase might fail cause of 429 error

```
   Changes to Outputs:
    - service_endpoints = [
        - {
            - host     = "cepkfq4dzz.westus3.azure.clickhouse.cloud"
            - port     = 9440
            - protocol = "nativesecure"
          },
        - {
            - host     = "cepkfq4dzz.westus3.azure.clickhouse.cloud"
            - port     = 8443
            - protocol = "https"
          },
      ] -> null
    - service_iam       = "iam-ch-production-westus3-burgundy-mi-58" -> null
  clickhouse_service.service: Destroying... [id=f3b067ce-7c32-4ea9-a0d1-1fcd1325d167]
  
  Error: Error Deleting ClickHouse Service
  
  Could not delete service, unexpected error: status: 429, body:
  {"error":"TOO_MANY_REQUESTS"}
```

This is a retryable error and should be handled by the api client.

Similarly we might handle 500s as well

whites11:
merged in main, will be released in 1.0.0

PR #143, opened 2024-08-22
Improve retry logic

whites11:
Towards: https://github.com/ClickHouse/terraform-provider-clickhouse/issues/133

This PR:

- Replaces own code with backoff library for backoff type use cases
- Add automatic retry of 429 and 5xx errors

as a side effect this PR makes end to end tests dramatically more reliable

### DIFF
--- a/pkg/internal/api/client.go
+++ b/pkg/internal/api/client.go
@@ -8,6 +8,8 @@ import (
 	"net/http"
 	"strings"
 	"time"
+
+	"github.com/cenkalti/backoff/v4"
 )
 
 type ClientImpl struct {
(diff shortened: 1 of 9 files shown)


pkg/internal/api/client.go
```diff
@@ -219,6 +227,33 @@ func (c *ClientImpl) CreateService(s Service) (*Service, string, error) {
 	return &serviceResponse.Result.Service, serviceResponse.Result.Password, nil
 }
 
+func (c *ClientImpl) WaitForServiceState(serviceId string, stateChecker func(string) bool, maxWaitSeconds int) error {
+	// Wait until service is in desired status
```
nikita-vanyasin, line 231:
```suggestion
	// Wait until service is in desired state
```

pkg/internal/api/client.go
```diff
@@ -219,6 +227,33 @@ func (c *ClientImpl) CreateService(s Service) (*Service, string, error) {
...
+	checkStatus := func() error {
```
nikita-vanyasin, line 232:
```suggestion
	checkState := func() error {
```

pkg/internal/api/client.go
```diff
@@ -219,6 +227,33 @@ func (c *ClientImpl) CreateService(s Service) (*Service, string, error) {
...
+		}
+
+		return fmt.Errorf("service %s is in state %s", serviceId, service.State)
+	}
+
+	if maxWaitSeconds < 5 {
+		maxWaitSeconds = 5
+	}
+
+	err := backoff.Retry(checkStatus, backoff.WithMaxRetries(backoff.NewConstantBackOff(5*time.Second), uint64(maxWaitSeconds/5)))
```
nikita-vanyasin, line 249:
```suggestion
	err := backoff.Retry(checkState, backoff.WithMaxRetries(backoff.NewConstantBackOff(5*time.Second), uint64(maxWaitSeconds/5)))
```

pkg/internal/api/client.go
```diff
@@ -300,27 +335,13 @@ func (c *ClientImpl) UpdateServicePassword(serviceId string, u ServicePasswordUp
...
-}
-
 func (c *ClientImpl) DeleteService(serviceId string) (*Service, error) {
 	service, err := c.GetService(serviceId)
 	if err != nil {
 		return nil, err
 	}
 
-	if service.State != "stopped" && service.State != "stopping" {
+	if service.State != StatusStopped && service.State != StatusStopping {
```
nikita-vanyasin, line 344: Does it make sense to rename these constants from StatusXXX to StateXXX? To make sure naming is consistent with what is shown in CP?

nikita-vanyasin (approved):

pkg/internal/api/client.go
whites11: ok will rename everything.

PR #144, opened 2024-08-23
Add user agent

whites11:
For tracking purposes, we want to add a user-agent header to all http requests to the API.
This way we can server side have an idea on how many users are using the terraform provider and which version

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -240,4 +240,3 @@ jobs:
           # GitHub sets the GITHUB_TOKEN secret automatically.
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GPG_FINGERPRINT: ${{ steps.import_gpg.outputs.fingerprint }}
-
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yaml
```diff
@@ -241,3 +247,33 @@ jobs:
...
+        uses: actions/checkout@v4
+        with:
+          ssh-key: ${{secrets.DEPLOY_SSH_PRIVATE_KEY }}
+      - name: Bump project version in go code
+        id: bump-dev
+        run: |
+          version="${{ inputs.version }}"
+          
+          # Bump one patch up and add '-dev' suffix
+          dev="$(echo "${version}" | awk -F. '{$NF = $NF + 1;} 1' | sed 's/ /./g')-dev"
```
nikita-vanyasin, line 269: Seems a bit hacky!
Maybe use ldflags for go build instead?

nikita-vanyasin: example: https://github.com/ClickHouse/data-plane-application/blob/main/dataplanectl/Makefile#L51

nikita-vanyasin: Also, worth adding git commit version IMO. With ldflags it is easier to pass multiple params than adding sed commands

whites11: Man, I have been doing go for years and never knew this! Thanks!

rndD (approved):
LGTM

PR #159, opened 2024-09-25
split api client functions into smaller files

whites11:
to simplify code a bit, I split the api client into smaller files grouped by type of call.

I also replaced a bunch of structs with a generic struct to remove some code


pkg/internal/api/common.go
```diff
@@ -0,0 +1,135 @@
...
+				currentExponentialBackoff = currentExponentialBackoff * 2
+
+				return nil, fmt.Errorf("status: %d, body: %s", res.StatusCode, body)
+			}
+
+			return body, nil
+		}
+	}
+
+	// This is a fake exponential backoff, becacuse multiplier is only 1.
```
nikita-vanyasin, line 121:
```suggestion
	// This is a fake exponential backoff, because multiplier is only 1.
```

pkg/internal/api/service.go
```diff
@@ -0,0 +1,197 @@
...
+type ServiceResponseResult struct {
+	Service  Service `json:"service"`
+	Password string  `json:"password"`
+}
+
+type ServiceBody struct {
+	Service Service `json:"service"`
+}
+
+// GetService - Returns a specifc order
```
nikita-vanyasin, line 27:
```suggestion
// GetService - Returns service by ID
```

nikita-vanyasin (approved):

PR #161, opened 2024-09-26
Improve validation

whites11:
Use terraform validators instead of custom ones where possible

### DIFF
--- a/pkg/resource/service.go
+++ b/pkg/resource/service.go
@@ -10,6 +10,9 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/hashicorp/terraform-plugin-framework-validators/stringvalidator"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+
 	"github.com/ClickHouse/terraform-provider-clickhouse/pkg/internal/api"
 	"github.com/ClickHouse/terraform-provider-clickhouse/pkg/resource/models"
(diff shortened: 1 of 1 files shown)


pkg/resource/service.go
```diff
@@ -65,20 +68,43 @@ func (r *ServiceResource) Schema(_ context.Context, _ resource.SchemaRequest, re
...
+					stringvalidator.AtLeastOneOf(path.Expressions{
+						path.MatchRoot("password_hash"),
+					}...),
+				},
 			},
 			"password_hash": schema.StringAttribute{
 				Description: "SHA256 hash of password for the default user. One of either `password` or `password_hash` must be specified.",
 				Optional:    true,
 				Sensitive:   true,
+				Validators: []validator.String{
```
nikita-vanyasin, line 82: Maybe go step further and actually try to decode base64 instead of validating against regex? I guess it will also work faster. (not sure if validator.String can have some custom validator passed)

nikita-vanyasin: Ah I see it was just moved from other place so it is not really related to PR.

nikita-vanyasin (approved):

pkg/resource/service.go
whites11: it's a good suggestion, I will keep it for a future round of improvements

Issue #184, opened 2024-11-01
Unable to use the provider to manage `dev` instances of Clickhouse after upgrading to provider version `1.2.4`, `num_replicas` not supported on `dev` instances


jamiezieziula:
After upgrading to provider version `1.2.4`, trying to apply `terraform` fails with the following error:
```bash
Error: Error Updating ClickHouse Service Scaling
with clickhouse_service.events
on clickhouse.tf line 42, in resource "clickhouse_service" "events":
resource "clickhouse_service" "events" {
Could not update service scaling, unexpected error: status: 400, body: {"requestId":"2b03ac68-e70b-4eb7-a0c5-4e6c54476de8","error":"BAD_REQUEST: request body.numReplicas: Number of replicas must be greater than 3 <number:0>","status":400}
```

As documented in the provider, `num_replicas` is only support on `production` instances. However, even when omitting the `num_replicas` configuration altogether, or passing `null` as the value, the provider still attempts to reconcile it.

nikita-vanyasin:
Hi @jamiezieziula
Thanks for the report! We will take a look and keep you updated 👌🏻

whites11:
hello @jamiezieziula and thanks for your report.
I tried replicating this by creating a new development tier service and I could not.
I assume the service you're witnessing this error on existed before 1.2.4 and the issue started only after the upgrade.

I also tried creating a dev service using version 1.2.3 and then upgrading to 1.2.4, still with no error.

Would it be possible to share the .tfstate file's content? we only need the `num_replicas` field of the `service` resource. Example:

```
"resources": [
    {
      "mode": "managed",
      "type": "clickhouse_service",
      "name": "service",
      "provider": "provider[\"registry.terraform.io/clickhouse/clickhouse\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            ...
            "num_replicas": null,
           ...
```

mjroeleveld:
I'm also getting this error

whites11:
> I'm also getting this error

same question for you then, can you please provide the details I requested above?

mjroeleveld:
`num_replicas` is `null` because it complains that you cannot provide it for development.
<img width="815" alt="Screenshot 2024-11-04 at 16 26 24" src="https://github.com/user-attachments/assets/eabf4d8b-54ca-4342-be12-80a23e957370">
[...]

PR #185, opened 2024-11-06
Update num_replicas when value is known

smstone:
Update `num_replicas` value on service only when `num_replicas` is not null and a known value.

Addresses #184

### DIFF
--- a/pkg/resource/service.go
+++ b/pkg/resource/service.go
@@ -114,7 +114,7 @@ func (r *ServiceResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 				Required:    true,
 			},
 			"tier": schema.StringAttribute{
-				Description: "Tier of the service: 'development', 'production'. Production services scale, Development are fixed size.",
+				Description: "Tier of the service: 'development', 'production'. Production services scale and development is a fixed size.",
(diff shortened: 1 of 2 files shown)


pkg/resource/service.go
```diff
@@ -793,7 +793,8 @@ func (r *ServiceResource) Update(ctx context.Context, req resource.UpdateRequest
 	}
 	if plan.NumReplicas != state.NumReplicas {
 		scalingChange = true
-		if !plan.NumReplicas.IsNull() {
+		// Only update `num_replicas` on production tier.
+		if !plan.NumReplicas.IsNull() && plan.Tier.ValueString() == api.TierProduction {
```
whites11, line 797: One day we might have a third tier or we might allow changing replicas for the dev tier.
I would prefer to keep num replicas editable in principle (a.k.a. not tied to a specific tier).

Why don't we simply extend the check as

```
if !plan.NumReplicas.IsNull() && !plan.NumReplicas.IsUnknown() {
```

this way we basically delegate num_replicas to the server side and we control which tiers can set the num_replicas by using validation like we do on line 362

whites11 (approved):

PR #199, opened 2024-12-03
add support for BYOC

whites11:
towards: https://github.com/ClickHouse/sre/issues/150

Add BYOC support


pkg/resource/models/service_resource.go
```diff
@@ -98,6 +98,7 @@ func (b BackupConfiguration) ObjectValue() basetypes.ObjectValue {
 
 type ServiceResourceModel struct {
 	ID                              types.String `tfsdk:"id"`
+	BYOCId                          types.String `tfsdk:"byoc_id"`
```
nikita-vanyasin, line 101:
```suggestion
	BYOCID                          types.String `tfsdk:"byoc_id"`
```

In same struct, there is `ID` field, IMHO makes sense to follow same style for other fields. (and for models.Service struct, but there is `Id` field).

nikita-vanyasin (approved):

pkg/resource/models/service_resource.go
whites11: fair point

PR #207, opened 2024-12-18
Resource database

whites11:
Towards: https://github.com/ClickHouse/sre/issues/176

this is first PR to allow managing databases, tables and users in a clickhouse service via terraform.

This PR specifically adds support for database.

NOTE: this is not mergeable yet, because it still requires clickops on a newly created service to enable query api on it. OpenAPI support to do this is not there yet.


nikita-vanyasin (approved):

pkg/internal/api/common.go
```diff
@@ -133,3 +136,28 @@ func (c *ClientImpl) doRequest(ctx context.Context, req *http.Request) ([]byte,
 
 	return body, err
 }
+
+func (c *ClientImpl) runQuery(ctx context.Context, serviceID string, sql string, args ...interface{}) ([]byte, error) {
+	qry, err := sqlbuilder.ClickHouse.Interpolate(sql, args)
+	if err != nil {
+		return nil, err
+	}
```
jkaflik, line 144: Is there any reason we don't leverage ClickHouse query parameters (they are available in API as per documentation: https://clickhouse.com/docs/en/get-started/query-endpoints#query-parameters ) and do query interpolation on the client-side?

pkg/internal/api/common.go
```diff
@@ -133,3 +136,28 @@ func (c *ClientImpl) doRequest(ctx context.Context, req *http.Request) ([]byte,
...
+
+	s := struct {
+		Sql string `json:"sql"`
```
jkaflik, line 147:
```suggestion
		SQL string `json:"sql"`
```

nit: practice is acronyms in names in Go should be the either full lower or full case (depend whether we expose outside package or not).

pkg/internal/api/common.go
```diff
@@ -133,3 +136,28 @@ func (c *ClientImpl) doRequest(ctx context.Context, req *http.Request) ([]byte,
...
+	}{
+		Sql: qry,
+	}
+
+	rb, err := json.Marshal(s)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequest(http.MethodPost, c.getQueryAPIPath(serviceID), strings.NewReader(string(rb)))
```
jkaflik, line 157:
```suggestion
	buffer := &bytes.Buffer{}
	if err := json.NewEncoder(buffer).Encode(&s); err != nil {
		return nil, fmt.Errorf("encoding query payload to JSON failed: %w", err)
	}
	req, err := http.NewRequest(http.MethodPost, c.getQueryAPIPath(serviceID), buffer)
```

nit: more straightforward to operate with io.Reader impl

pkg/internal/api/common.go
jkaflik: I guess ClickHouse can miss query parameters parser support for some of the statements.

whites11: Yes exactly. Not all fields are supported as parameters and thus I prefer having a uniform query handling on the client side rather than an hybrid approach where some params are handled by the server and some by the client

nikita-vanyasin (approved):

PR #211, opened 2025-01-14
PPv2 support

whites11:
Add support for ppv2.

There is some naming to be checked in the `tier` field description, but the logic is ready to be reviewed


nikita-vanyasin (approved):

nikita-vanyasin:
@whites11 given that with this PR merged we gonna have 2.0 version, maybe it makes sense to update main README in the same PR to show changes between v1 and v2 (upgrade instructions).

PR #213, opened 2025-01-14
Hydra

whites11:
Towards: https://github.com/ClickHouse/sre/issues/255

add support for compute-compute separation services


docs/resources/service.md
```diff
@@ -4,12 +4,19 @@ page_title: "clickhouse_service Resource - clickhouse"
 subcategory: ""
 description: |-
   You can use the clickhouse_service resource to deploy ClickHouse cloud instances on supported cloud providers.
+  Known limitations:
+  If you create a service with warehouse_id set and then remove warehouse_id attribute completely, the provider won't detect the change. If you want to make a secondary service become primary, remove the warehouse_id and taint it before applying.If you create a service with readonly flag set to true and then remove readonly flag completely, the provider won't detect the change. If you want to make a secondary service read write, explicitely set the readonly flag to false.
```
nikita-vanyasin, line 8:
```suggestion
  If you create a service with warehouse_id set and then remove warehouse_id attribute completely, the provider won't detect the change. If you want to make a secondary service become primary, remove the warehouse_id and taint it before applying. If you create a service with readonly flag set to true and then remove readonly flag completely, the provider won't detect the change. If you want to make a secondary service read write, explicitly set the readonly flag to false.
```

nikita-vanyasin (approved):
LGTM. 
One nit: 'readonly' field name seems to be quite broad.
I can imagine a situation in future when we allow to switch the non-Hydra service into read-only mode. Will we be able to re-use this field?

whites11:
> LGTM. One nit: 'readonly' field name seems to be quite broad. I can imagine a situation in future when we allow to switch the non-Hydra service into read-only mode. Will we be able to re-use this field?

definitely. should that be needed in the future, I wouldn't like to have two separate fields with the same meaning.

PR #214, opened 2025-01-16
Experimental ClickPipe support

jkaflik:
This PR adds experimental ClickPipe resource support.
Known limitations & bugs are listed in the auto-generated document.

This feature will be released as a pre-release version. It requires Cloud organisation have a feature flag enabled.


examples/full/clickpipe/kafka_azure_eventhub/main.tf
```diff
@@ -0,0 +1,74 @@
...
+variable "azure_eventhub_connection_string" {
+  description = "Connection string for Azure EventHub"
+  sensitive   = true
+}
+
+resource "clickhouse_clickpipe" "kafka_azure_eventhub" {
+  name        = "Azure EventHub 🚀 ClickPipe"
+  description = "Data pipeline from Azure EventHub to ClickHouse"
+
+  service_id = var.service_id
```
whites11, line 22: examples in the `examples/full` directory are meant to be self sufficient, meaning people should be able to just set the variables for authentication and `terraform apply` the example. In this case, the service is not created in the example.
Please note examples in `examples/full` are used by e2e tests and thus this test would never pass.

examples/full/clickpipe/kafka_azure_eventhub/provider.tf
```diff
@@ -0,0 +1,23 @@
+terraform {
```
whites11, line 1: Like we do in the other examples, please add a template file for the provider (see https://github.com/ClickHouse/terraform-provider-clickhouse/blob/clickpipe/examples/full/development/aws/provider.tf.template) this will allow automation to automatically bump examples to use latest release as part of the release automation

pkg/resource/clickpipe.go
```diff
@@ -0,0 +1,1092 @@
...
+				},
+				Optional: true,
+			},
+			"state": schema.StringAttribute{
+				MarkdownDescription: "The state of the ClickPipe. (`Running`, `Stopped`). Default is `Running`. Whenever the pipe state changes, the Terraform provider will try to ensure the actual state matches the planned value. If pipe is `Failed` and plan is `Running`, the provider will try to resume the pipe. If plan is `Stopped`, the provider will try to stop the pipe. If the pipe is `InternalError`, no action will be taken.",
+				Optional:            true,
+				Default:             stringdefault.StaticString(api.ClickPipeRunningState),
+				Computed:            true,
+				Validators: []validator.String{
+					stringvalidator.OneOf(api.ClickPipeRunningState, api.ClickPipeStoppedState),
```
whites11, line 137: this confuses me a bit. you say valid values are "Running" and "Stopped" but then in the description you mention "Failed" and "InternalError" as well. Does it work at all?
What I mean, if the server pushes value "Failed" to the terraform client, will that be validated when storing the state or will the validation only apply to client-supplied  plan?

pkg/resource/clickpipe.go
```diff
@@ -0,0 +1,1092 @@
...
+						Description: "The name of the ClickHouse table.",
+						Required:    true,
+					},
+					"managed_table": schema.BoolAttribute{
+						MarkdownDescription: "Whether the table is managed by ClickHouse Cloud. If `false`, the table must exist in the database. Default is `true`.",
+						Default:             booldefault.StaticBool(true),
+						Computed:            true,
+						Optional:            true,
+					},
+					"table_definition": schema.SingleNestedAttribute{
```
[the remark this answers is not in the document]
jkaflik: This will not be a problem in the future. This will be dropped feature once we have table management in Terraform provider. ;)

pkg/resource/clickpipe.go
jkaflik: Validator matches against user input. Yes, we expect the user to provide `Running` or `Failed` only.
However, ClickPipe can have various states, like `Failed`.  I've tested it with Terraform state, and it works fine.
Whenever API returns other status, Terraform consumes it and proposes update action accordingly. 

I will make the description a bit more descriptive.

pkg/resource/models/clickpipe_resource.go
```diff
@@ -0,0 +1,392 @@
+package models
+
+import (
+	"github.com/hashicorp/terraform-plugin-framework/attr"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+)
+
+type ClickPipeScalingModel struct {
+	Replicas types.Int64 `tfsdk:"replicas"`
+	// Concurrency types.Int64 `tfsdk:"concurrency"`
```
whites11, line 10: please remove this line if not needed thanks

pkg/resource/models/clickpipe_resource.go
```diff
@@ -0,0 +1,392 @@
...
+}
+
+func (m ClickPipeScalingModel) ObjectType() types.ObjectType {
```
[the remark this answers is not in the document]
jkaflik: It's a bit hard to deal with some aspects of the code. It's more DX compared to performance. Let's keep it like this.

whites11 (approved):

PR #219, opened 2025-01-24
Alpha builds

whites11:
We have an experimental resource in `main`: "clickpipe".

We want to make that resource available in pre releases, but not in stable ones.

For example, version 2.0.0 should NOT contain the `clickpipe` resource while 2.0.0-alpha1 should contain it.

In order to do so, this PR makes the following changes:

- add the `alpha` [go build tag](https://www.digitalocean.com/community/tutorials/customizing-go-binaries-with-build-tags) to all the go files related to the clickpipe resource.
- The above required a refactoring of the provider's initialization files.
- have separate configs for goreleaser 
- call different make targets generating docs during a release
- avoid bumping examples for pre releases
- use custom version of tfplugindocs to add support for go build tags


.goreleaser-alpha.yml
```diff
@@ -0,0 +1,63 @@
...
+      - darwin
+    goarch:
+      - amd64
+      - '386'
+      - arm
+      - arm64
+    ignore:
+      - goos: darwin
+        goarch: '386'
+    tags:
```
whites11, line 32: this is a copy of the original config file, with only this section changed

nikita-vanyasin: Maybe let's add this comment right into file header? Otherwise it is not quite clear why we having two almost identical files. 

By the way, I thought it was possible with goreleaser to have 2 build targets inside one file. Then you can use Yaml anchors/aliases to deduplicate

nikita-vanyasin: Feel free to merge this anyway, if we need it before 2.0.0 release. We can always de-duplicate later

jkaflik (approved):
Looks good 🚀

.goreleaser-alpha.yml
whites11: it requires goreleaser pro license

PR #221, opened 2025-01-28
Experimental ClickPipe resource fixes

jkaflik:
- remove import implementation as it's broken now
- Omit empty timestamp in Kafka offset strategy
- Wait desired plan state is reached after scaling ClickPipe
- Restore variables in ClickPipes provider examples
- Add a UseStateForUnknown modifier for Kafka consumer group attribute
- Allow empty description
- Handle ClickPipe state changes more gracefully
- multiple_pipes_example

### DIFF
--- a/examples/clickpipe/kafka_azure_eventhub/provider.tf
+++ b/examples/clickpipe/kafka_azure_eventhub/provider.tf
@@ -8,8 +8,16 @@ terraform {
   }
 }
 
+variable "organization_id" {}
+variable "token_key" {}
+variable "token_secret" {}
+variable "api_url" {
+  default = "https://api.clickhouse.cloud/v1"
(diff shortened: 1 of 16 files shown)


examples/clickpipe/kafka_azure_eventhub/provider.tf
```diff
@@ -8,8 +8,16 @@ terraform {
   }
 }
 
+variable "organization_id" {}
+variable "token_key" {}
+variable "token_secret" {}
+variable "api_url" {
+  default = "https://api.clickhouse.cloud/v1"
```
whites11, line 15: this is already the default on the client API side, why are you defaulting it again?

whites11 (approved):

PR #228, opened 2025-02-07
add user resource

whites11:
Towards: https://github.com/ClickHouse/sre/issues/307

this PR adds a new alpha resource to allow creating a user in a clickhouse service.

This is on purpose a partial PR: it is possible to create and update and delete users, but there are no roles assigned so users created by this resource are currently useless.

Managing and assignment of roles will come in a future PR

### DIFF
--- a/pkg/internal/api/db_user.go
+++ b/pkg/internal/api/db_user.go
@@ -1,0 +1,90 @@
+package api
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/huandu/go-sqlbuilder"
+
+	sqlutil "github.com/ClickHouse/terraform-provider-clickhouse/pkg/internal/sql"
+)
+
+type User struct {
+	Name               string `json:"name"`
+	PasswordSha256Hash string `json:"-"`
+}
+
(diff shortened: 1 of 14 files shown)


pkg/internal/api/db_user.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	// We don't get the SHA back from clickhouse, so we assume the operation was successful and return it to the client.
+	createdUser.PasswordSha256Hash = user.PasswordSha256Hash
+
+	return createdUser, nil
+}
+
+func (c *ClientImpl) GetUser(ctx context.Context, serviceID string, name string) (*User, error) {
+	format := "SELECT name FROM system.users WHERE name = ${name} and storage = 'replicated'"
```
nikita-vanyasin, line 45: Can you please add inline comment why storage is 'replicated'?

nikita-vanyasin (approved):

pkg/internal/api/db_user.go
whites11: done

PR #229, opened 2025-02-10
Make Kafka credentials optional when working with IAM role authentication method

jkaflik:
This changes attribute to optional and changes and logic execution order to make sure the proper plan and API request is executed.
Also fixes a missing IAM role and CA certificate in API call.


whites11 (approved):

pkg/resource/clickpipe.go
```diff
@@ -541,45 +536,63 @@ func (c *ClickPipeResource) Create(ctx context.Context, request resource.CreateR
...
+		}
+
+		if kafkaModel.Authentication.ValueString() != api.ClickPipeAuthenticationIAMRole {
+			if !kafkaModel.Credentials.IsNull() {
+				credentialsModel := models.ClickPipeKafkaSourceCredentialsModel{}
+				response.Diagnostics.Append(kafkaModel.Credentials.As(ctx, &credentialsModel, basetypes.ObjectAsOptions{})...)
+
+				var credentials *api.ClickPipeKafkaSourceCredentials
+
+				if kafkaModel.Authentication.ValueString() != api.ClickPipeAuthenticationIAMRole {
```
kitop, line 564: This is the same we're checking a few lines above. Can probably get rid of the check as it will always be true.

PR #238, opened 2025-02-19
Avoid marking alpha releases as "pre releases"

whites11:
Marking alpha releases as "pre release" prevent them from being picked up by hashicorp in the public terraform provider page.
We still don't mark the alpha releases as "latest" so it shouldn't change much on github side.


jkaflik (approved):
LGTM. Eventually we can add additional GH API call to mark it as pre-release with a 5 second delay so Hashicorp is able to pick up release.

Issue #69, opened 2024-06-25
Make it possible to enable MySQL protocol


incfly:
https://registry.terraform.io/providers/ClickHouse/clickhouse/latest/docs/resources/service 

We have double SHA1 password but not seeing an option to enable MySQL. We might need control plane to add rest api for this. cc @Kinzeng 

I encountered this when was working on some internal testing automation.  https://github.com/ClickHouse/data-plane-application/issues/10943

nikita-vanyasin:
Blocked by https://github.com/ClickHouse/control-plane/issues/9964

serdec:
@incfly @nikita-vanyasin how do you envision the user to setup a username and password for mysql?

noamliber-reco:
Very important request

whites11:
> Very important request

Still blocked by https://github.com/ClickHouse/control-plane/issues/9964

nothing we can do about it

whites11:
@noamliber-reco I realised I linked to a private ticket, sorry about that.
We have an internal dependency that is preventing us from fulfilling your request, but it's in our roadmap.
Thanks for your patience

whites11:
released with terraform provider version 2.1.0

PR #241, opened 2025-02-20
Add Mysql endpoint support

whites11:
Towards: https://github.com/ClickHouse/terraform-provider-clickhouse/issues/69

Added new attribute to `clickhouse_service` to allow enabling the mysql protocol in the service


pkg/internal/api/models.go
```diff
@@ -1,5 +1,9 @@
 package api
 
+const (
+	PROTOCOL_MYSQL = "mysql"
```
nikita-vanyasin, line 4:
```suggestion
	EndpointProtocolMySQL = "mysql"
```

pkg/resource/models/service_resource.go
```diff
@@ -72,6 +72,46 @@ func (p PrivateEndpointConfig) ObjectValue() basetypes.ObjectValue {
 	})
 }
 
+type EndpointsConfiguration struct {
+	Mysql types.Object `tfsdk:"mysql"`
```
nikita-vanyasin, line 76:
```suggestion
	MySQL types.Object `tfsdk:"mysql"`
```

nikita-vanyasin (approved):

PR #283, opened 2025-04-17
Mysql plan

whites11:
Towards: https://github.com/ClickHouse/sre/issues/382

smarter ModifyPlan to avoid marking the `mysql` endpoint as always dirty.

Related to support escalation https://github.com/ClickHouse/support-escalation/issues/4509

### DIFF
--- a/pkg/resource/service.go
+++ b/pkg/resource/service.go
@@ -724,12 +724,10 @@ func (r *ServiceResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
 
 			resp.Plan.Set(ctx, plan)
 		}
-
 	} else {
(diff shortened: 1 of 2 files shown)


pkg/resource/service.go
```diff
@@ -1530,10 +1563,6 @@ func (r *ServiceResource) UpgradeState(ctx context.Context) map[int64]resource.S
 							},
 						},
 					},
-					"has_tranparent_data_encryption": schema.BoolAttribute{
```
nikita-vanyasin, line 1533: TDE removal does not seem related to this PR?

nikita-vanyasin (approved):

PR #286, opened 2025-04-23
private endpoint refactoring

whites11:
Towards: https://github.com/ClickHouse/sre/issues/336

This PR reworks how private endpoint configuration can be done with terraform.
Refer for the changes in the README.md file for details how this is changed.


README.md
```diff
@@ -14,6 +14,67 @@ Please refer to the [official docs](https://registry.terraform.io/providers/Clic
 
 ## Breaking changes and deprecations
 
+### Upgrading to version >= 3.2.0
+
+In version 3.2.0 we introduced a change in the `Private Endpoints` feature that requires a change on your side if you use this setting.
```
nikita-vanyasin, line 19: should it be v4 if it requires an action from the user?

nikita-vanyasin (approved):

README.md
whites11: it doesn't strictly require an action.
users can use the old config and it'll keep working as-is with no modifications shown.
but they are encouraged to upgrade to the new system and if they try to create resources "the old way" they are blocked.

whites11: also I don't want to make too many major releases in a row :sweat_smile:

PR #302, opened 2025-05-14
add the capability to create new services from existing backups

egor-krv:
Added support to restore from a backup when creating a new resource.

When calling

```
POST /v1/organizations/{organizationId}/services
```

to create a new service, allowed passing an optional field

```
"backupId": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
```

Ensured:

- [x] It works with unencrypted backups
- [x] It works with encrypted/TDE backups
- [x] The field can be left out
- [x] The field can be set null
- [x] The field canNOT be set to the empty string
- [x] The field cannot be changed or removed from the resource

### DIFF
--- a/pkg/resource/service.go
+++ b/pkg/resource/service.go
@@ -479,7 +479,7 @@ func (r *ServiceResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
 			)
 		}
 
-		if !plan.BackupID.IsNull() && !plan.BackupID.IsUnknown() && plan.BackupID != state.BackupID {
+		if plan.BackupID != state.BackupID {
 			resp.Diagnostics.AddAttributeError(
 				path.Root("backup_id"),
 				"Invalid Update",
@@ -974,8 +974,6 @@ func (r *ServiceResource) Create(ctx context.Context, req resource.CreateRequest
 		)
 		return
 	}
-	// Always rewrite the backup ID as we need to make Terraform state to work by API does not return it backup ID
-	s.BackupID = service.BackupID
 
 	err = r.client.WaitForServiceState(ctx, s.Id, func(state string) bool { return state != api.StateProvisioning }, 20*60)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/resource/service.go
```diff
@@ -73,6 +73,13 @@ func (r *ServiceResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 					stringplanmodifier.UseStateForUnknown(),
 				},
 			},
+			"backup_id": schema.StringAttribute{
+				Description: "ID of the backup to restore when creating new service. If specified, the service will be created as a restore operation",
+				Optional:    true,
+				PlanModifiers: []planmodifier.String{
```
whites11, line 79: Since you block modification in ModifyPlan you don't need RequiresReplace() here.

pkg/resource/service.go
```diff
@@ -472,6 +479,22 @@ func (r *ServiceResource) ModifyPlan(ctx context.Context, req resource.ModifyPla
 			)
 		}
 
+		if !plan.BackupID.IsNull() && !plan.BackupID.IsUnknown() && plan.BackupID != state.BackupID {
```
whites11, line 482: I think in this case `plan.BackupID != state.backupID` is enough for both scenarios here.
if either is null and the other is not, it will catch it.
If they're both non-null but different, it'll catch it.

am I wrong?

pkg/resource/service.go
```diff
@@ -1684,6 +1713,7 @@ func (r *ServiceResource) UpgradeState(ctx context.Context) map[int64]resource.S
 					EncryptionAssumedRoleIdentifier types.String `tfsdk:"encryption_assumed_role_identifier"`
 					QueryAPIEndpoints               types.Object `tfsdk:"query_api_endpoints"`
 					BackupConfiguration             types.Object `tfsdk:"backup_configuration"`
+					BackupID                        types.String `tfsdk:"backup_id"`
```
whites11, line 1716: this struct represents a Service in an older version of the provider, where BackupID was not defined so this field defined here is wrong.
You have to remove it and define it in the sruct on line 1778 as a null object. Similarly to what I did for TransparentEncryptionData

pkg/resource/service.go
egor-krv: I can see other fields like "if !plan.BYOCID.IsNull() && !plan.BYOCID.IsUnknown()" are done this way. I assumed that I have to use methods provided to compare. I updated the code as per your suggestion.

pkg/resource/service.go
egor-krv: Updated

whites11 (approved):

PR #317, opened 2025-06-17
Disable RBAC resources from the alpha build

whites11:
We switched to the external provider for the same feature.
Proper cleanup will follow


rsickles:
LGTM

nikita-vanyasin (approved):

rsickles (approved):
LGTM

Issue #325, opened 2025-07-14
Add support for "slow" release channel


danxmoran:
I tried to switch one of our services to the slow channel, and got the error:
```
│ Error: Invalid Attribute Value Match
│
│   with module.service.clickhouse_service.service,
│   on ../../../../../modules/clickhouse/service/service.tf line 17, in resource "clickhouse_service" "service":
│   17:   release_channel = var.use_slow_channel ? "slow" : "default"
│
│ Attribute release_channel value must be one of: ["default" "fast"], got: "slow"
```

whites11:
version 3.4.0 is out and should allow slow release channel. I will go on and close the ticket, feel free to reopen if you encounter problems.

PR #327, opened 2025-07-15
add support for slow release channel

whites11:
towards: https://github.com/ClickHouse/terraform-provider-clickhouse/issues/325

### DIFF
--- a/examples/full/basic/aws/main.tf
+++ b/examples/full/basic/aws/main.tf
@@ -24,8 +24,8 @@ variable "release_channel" {
   type = string
   default = "default"
   validation {
-    condition     = var.release_channel == "default" || var.release_channel == "fast"
-    error_message = "Release channel can be either 'default' or 'fast'."
+    condition     = var.release_channel == "default" || var.release_channel == "fast" || var.release_channel == "slow"
+    error_message = "Release channel can be 'default', 'fast' or 'slow'."
   }
 }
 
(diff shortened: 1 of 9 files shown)


examples/full/basic/aws/main.tf
```diff
@@ -24,8 +24,8 @@ variable "release_channel" {
   type = string
   default = "default"
   validation {
-    condition     = var.release_channel == "default" || var.release_channel == "fast"
-    error_message = "Release channel can be either 'default' or 'fast'."
+    condition     = var.release_channel == "default" || var.release_channel == "fast" || var.release_channel == "slow"
```
GMartinez-Sisti, line 27: nit for cleaner
```suggestion
    condition     = contains(["default", "fast", "slow"], var.release_channel)
```

pkg/internal/api/constants.go
```diff
@@ -7,6 +7,7 @@ const (
 
 	ReleaseChannelDefault = "default"
 	ReleaseChannelFast    = "fast"
+	ReleaseChannelSlow    = "slow"
```
GMartinez-Sisti, line 10: Not sure how valuable this is now, and might be out of scope for this PR but I see values in approaches like this to make sure we only accept the defined values:

```go
type ReleaseChannel string

const (
    ReleaseChannelDefault ReleaseChannel = "default"
    ReleaseChannelFast    ReleaseChannel = "fast"
    ReleaseChannelSlow    ReleaseChannel = "slow"
)
```

GMartinez-Sisti (approved):
LGTM with nits.

PR #331, opened 2025-07-18
Remove description field from clickpipes tf modules and examples

tpanetti:
This removes the description field from the clickpipes terraform modules and example since we have since removed it from ClickPipes and it causes the creation/update of ClickPipes to fail.

### DIFF
--- a/examples/clickpipe/externally_managed_table/main.tf
+++ b/examples/clickpipe/externally_managed_table/main.tf
@@ -3,34 +3,27 @@ variable "token_key" {}
 variable "token_secret" {}
 
 variable "service_id" {
-  description = "ClickHouse service ID"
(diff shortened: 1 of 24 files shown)


tsolodov (commented):
Looks like this field was already removed on ClickPipes side.  Ideally we should introduce deprecation of fields in advance. 

From customer UX, it's a breaking change. @whites11 , please chime in.

EvandroLG (approved):
LGTM!
Let’s go ahead with this change since it’s already been removed from ClickPipes and is now affecting Terraform users. We checked the data, and only a small number of pipes were created with the description field anyway.
Next time, we'll make sure to deprecate the field first before removing it :)

PR #332, opened 2025-07-24
Add Azure Blob Storage support to ClickPipe object storage

EvandroLG:
This PR adds support for Azure Blob Storage as a source for ClickPipe data ingestion pipelines, enabling users to ingest data directly from Azure Blob containers.

### Changes:
- **New object storage type:** `azureblobstorage`
- **New authentication method:** `CONNECTION_STRING` for Azure
- **New configuration fields:**
   - `connection_string` - Azure connection string (sensitive)
   - `path` - File path within container with wildcard support
   - `azure_container_name` - Azure container nam
   
Related to https://github.com/ClickHouse/clickpipes-platform/issues/4382

### DIFF
--- a/examples/clickpipe/object_storage_azure_blob/README.md
+++ b/examples/clickpipe/object_storage_azure_blob/README.md
@@ -1,0 +1,31 @@
+## ClickPipe Object Storage with Azure Blob Storage example
+
+This example demonstrates how to deploy a ClickPipe with an Azure Blob Storage container as the input source,
+authenticated with a connection string.
+
+## Prerequisites
+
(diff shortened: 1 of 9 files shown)


GMartinez-Sisti:
Should we name the folder `ClickPipes` instead?

jkaflik (approved):

GMartinez-Sisti:
> @GMartinez-Sisti let's handle the folder rename in a follow-up PR? Just to keep this one clean.

I was under the impression that the folder was being created here. Sorry for not checking beforehand!

PR #343, opened 2025-08-07
fix example for importing reverse private endpoint

taobojlen:
I noticed that in your docs for [reverse private endpoints](https://registry.terraform.io/providers/ClickHouse/clickhouse/3.4.1-alpha1/docs/resources/clickpipes_reverse_private_endpoint#import), it says that you can import an existing resource with

```
# ClickPipes reverse private endpoints can be imported by specifying both service ID and clickpipe ID.
terraform import clickhouse_clickpipes_reverse_private_endpoint.example xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx:xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
```

But this doesn't actually work! You need to specify the ID of the reverse private endpoint, not the ClickPipe that it's used in. Note that the ID of a reverse private endpoint isn't actually accessible via the UI; I had to use the API to find the ID.

I'm not sure how to update your actual Terraform docs but this is a start.


jkaflik:
Hey @taobojlen 

Thanks for your contribution. Documentation is auto-generated from this snippet.

That's true for ID missing in UI. We are currently working on UI improvements and this will be addressed as well.

tsolodov (approved):

PR #346, opened 2025-08-12
Fix issue where clickpipe scaling was failing due to type misalignment

### LINKED ISSUE #344: (3.6.0-alpha1) replicaCpuMillicores is failing type unmarshaling
existing clickpipe:

```hcl
resource "clickhouse_clickpipe" "traffic_logs" {
  name       = "traffic-logs-enriched"
  service_id = clickhouse_service.service.id

  scaling = {
    replicas = local.clickpipe_scaling_config[local.environment].replicas
  }
}
```

run a terraform plan

```
│ Could not read ClickPipe, unexpected error: json: cannot unmarshal string into Go struct field ClickPipeScaling.result.scaling.replicaCpuMillicores of type int64
```

no matter what value we put for `replica_cpu_millicores` (or leave it blank), the value from the API seems to be a string and the provider model expects an int64

tpanetti:
This fixes an issue where replica scaling for clickpipes would fail the terraform apply step (although it would actually apply), due to validation when the API would return back the kubernetes resource value which differed from the numeric input. This was not caught in testing due to the mock-clickpipes service implemented in control-plane returning those values as well.

Note there is still an issue where vertical scaling is not applying correctly. Still investigating this but wanted to get a patch out to restore replica scaling functionality.

This fixes #344


pkg/internal/api/clickpipe_models.go
```diff
@@ -1,11 +1,79 @@
...
+			}
+		} else if strings.HasSuffix(v, "M") {
+			// Megabytes - convert to GB (1000 M = 1 G)
+			str := strings.TrimSuffix(v, "M")
+			megabytes, parseErr := strconv.ParseFloat(str, 64)
+			if parseErr != nil {
+				_ = parseErr
+			} else {
+				val = megabytes / 1000.0 // Convert MB to GB
+			}
```
tsolodov, line 65: do we want to support `MB`/`GB` as well ? Might be easier to use parser from k8s https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource

tpanetti: Yeah didn't know if we were opposed to pulling in new libraries, much easier to use that. Added.

PR #361, opened 2025-08-29
fix(ci): add missing Go setup step in release workflow

jkaflik:
replaces https://github.com/ClickHouse/terraform-provider-clickhouse/pull/359


.github/workflows/release.yaml
```diff
@@ -324,8 +324,15 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
+          ref: "main"
```
jkaflik, line 327: This ensures we bump examples and regenerate docs in main always.

GMartinez-Sisti: This workflow can run from another branch, do we want to filter this to main? Not completely aware of this workflow yet.

PR #367, opened 2025-09-05
chore(azure): add private link example

GMartinez-Sisti:
# Why

- Add tests that require azure infrastructure

# What

- Add azure login
- Move duplicate into shared actions
- Lint

# Test

Deployed the private link example against dev, then created an instance on the same vnet and tested clickhouse connection:

```sh
azureuser@test-002:~$ curl --user 'default:xxxxxxxx' --data-binary 'SELECT 1' https://xxxxxxxxxx.westus3.privatelink.azure.clickhouse-dev.com:8443
1
```

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -178,7 +178,7 @@ jobs:
   upgrade:
     outputs:
       status: ${{ steps.status.outputs.status }}
-    needs: [ "validate", "token", "find-tf-releases", "list-examples" ]
+    needs: [ "validate", "token", "find-tf-releases", "list-examples", "e2e", "import" ]
     runs-on: k8s-large
     environment: default
     continue-on-error: true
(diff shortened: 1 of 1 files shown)


.github/actions/cleanup-cloud/action.yaml
```diff
@@ -0,0 +1,60 @@
+name: Cleanup AWS leftovers
```
nikita-vanyasin, line 1:
```suggestion
name: Cleanup CSP leftovers
```

.github/actions/cleanup-cloud/action.yaml
```diff
@@ -0,0 +1,60 @@
...
+description: Cleanup any AWS leftovers of e2e test
```
nikita-vanyasin, line 2:
```suggestion
description: Cleanup any CSP leftovers of e2e test
```

nikita-vanyasin (approved):

.github/actions/cleanup-cloud/action.yaml
GMartinez-Sisti: Nice catch!

PR #369, opened 2025-09-10
Add advanced settings for ClickPipes to terraform create/update

tpanetti:
# Overview
This adds support for settings objects on the ClickPipe terraform module. It provides support by doing the following
* Defining the settings type for POSTing new clickpipes
* Defining the GET and PUT endpoints for the new settings endpoints for clickpipes (GET is currently unused but defining for future usage).
* Set the flow so that creating the clickpipe will simply include the settings on the clickpipe but updating will instead hit the PUT endpoint

## Miscelaneous
* Also fixes an issue where the scaling object was being defined as empty rather than being undefined, this caused it to be included in the terraform plan when it should be optional.

## Notes
This relies on the OpenAPI changes which are projected to land on Thursday Sep 17th, so this will be merged then.


pkg/resource/clickpipe.go
```diff
@@ -656,6 +657,10 @@ func (c *ClickPipeResource) Schema(_ context.Context, _ resource.SchemaRequest,
 					},
 				},
 			},
+			"settings": schema.DynamicAttribute{
+				Description: "Advanced configuration options for the ClickPipe. These settings are specific to each pipe. For the complete list of available options, see the [OpenAPI documentation](https://clickhouse.com/docs/cloud/manage/api/swagger)",
+				Optional:    true,
+			},
```
jkaflik, line 663: What is our motivation for this vs typed structure in Terraform? I am not saying it's bad and probably might be viable approach, but I wonder if we considered other way and what are pros/cons.

tpanetti: > What is our motivation for this vs typed structure in Terraform

The major drive for this is to allow for dynamic keys here. If we went for the strongly typed route we would need to maintain the settings in this package which would result in another place where we would need to update the settings. Instead we can rely on the validation of OpenAPI for the settings and keep this package zero-touch. 

> For available options, would be great to point to a specific doc piece

Yes, planning on adding the proper link to the settings endpoint once it's deployed in prod on Thursday (:

gkorolev (approved):

PR #373, opened 2025-09-14
Add support for new table engines for ClickPipes managed table

tpanetti:
# Overview
This adds support for 3 new table engines in ClickPipe's managed table implementation (ReplacingMergeTree, SummingMergeTree and Null)

## Notes
Straightforward change for the most part, validation is properly done on OpenAPI with helpful error messages. Only thing to note is that VersionColumnID is not returned by the API so we maintain the state here.

### DIFF
--- a/pkg/resource/clickpipe.go
+++ b/pkg/resource/clickpipe.go
@@ -587,11 +587,20 @@ func (c *ClickPipeResource) Schema(_ context.Context, _ resource.SchemaRequest, 
 								Required:            true,
 								Attributes: map[string]schema.Attribute{
 									"type": schema.StringAttribute{
-										MarkdownDescription: "The type of the engine. Only `MergeTree` is supported.",
+										MarkdownDescription: "The type of the engine. Supported engines: `MergeTree`, `ReplacingMergeTree`, `SummingMergeTree`, `Null`.",
 										Required:            true,
 										Validators: []validator.String{
-											stringvalidator.OneOf("MergeTree"),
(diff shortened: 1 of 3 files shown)


pkg/resource/clickpipe.go
```diff
@@ -587,12 +587,21 @@ func (c *ClickPipeResource) Schema(_ context.Context, _ resource.SchemaRequest,
...
-											stringvalidator.OneOf("MergeTree"),
+											stringvalidator.OneOf("MergeTree", "ReplacingMergeTree", "SummingMergeTree", "Null"),
 										},
 									},
+									"version_column_id": schema.StringAttribute{
+										MarkdownDescription: "Column ID to use as version for ReplacingMergeTree engine. Required when engine type is `ReplacingMergeTree`.",
+										Optional:            true,
+									},
+									"column_ids": schema.ListAttribute{
+										MarkdownDescription: "Column IDs to sum for SummingMergeTree engine. Required when engine type is `SummingMergeTree`.",
```
whites11, line 601: I think we need to enforce the requirement check on the client side.
You could add a validator for "AlsoRequires" and "ConflictsWith" WDYT?

whites11 (approved):
looks ok, but let's please add validation about the fields that required based on the engine type

pkg/resource/clickpipe.go
tpanetti: Yeah makes sense to me, I'll add the validation here as well.

gkorolev (approved):

PR #374, opened 2025-09-15
Avoid confusing references to AWS red

JD557:
While looking at the private endpoint examples, I noticed some references to an `aws_red` service that doesn't seem to match the service name in other parts of the examples.

I'm assuming that this was a bad copy paste, so I made the examples consistent.

Note that I didn't check this and I'm assuming that's how the private service definition works (as I'm still learning about them). Please double check if the changes are correct 😅


tsolodov (approved):
Thank for the PR!

PR #388, opened 2025-10-02
WIP: support SQS Queue URL for continuous ingestion in S3

EvandroLG:
WIP

### DIFF
--- a/pkg/resource/clickpipe.go
+++ b/pkg/resource/clickpipe.go
@@ -5,6 +5,7 @@ package resource
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


examples/clickpipe/object_storage_s3_sqs_iam_role/provider.tf
```diff
@@ -0,0 +1,13 @@
+terraform {
+  required_providers {
+    clickhouse = {
+      source = "ClickHouse/clickhouse"
```
whites11, line 4: let's please use the template like we do here: https://github.com/ClickHouse/terraform-provider-clickhouse/blob/main/examples/clickpipe/externally_managed_table/provider.tf.template.alpha

pkg/resource/clickpipe.go
```diff
@@ -796,6 +810,44 @@ func (c *ClickPipeResource) ModifyPlan(ctx context.Context, request resource.Mod
...
+	if !plan.Source.IsNull() {
+		sourceModel := models.ClickPipeSourceModel{}
+		response.Diagnostics.Append(plan.Source.As(ctx, &sourceModel, basetypes.ObjectAsOptions{})...)
+
+		if !sourceModel.ObjectStorage.IsNull() {
+			objectStorageModel := models.ClickPipeObjectStorageSourceModel{}
+			response.Diagnostics.Append(sourceModel.ObjectStorage.As(ctx, &objectStorageModel, basetypes.ObjectAsOptions{})...)
+
+			// Validate queue_url is only provided when is_continuous is true
+			if !objectStorageModel.QueueURL.IsNull() && !objectStorageModel.QueueURL.IsUnknown() && objectStorageModel.QueueURL.ValueString() != "" {
```
whites11, line 823: you can and imho should do these checks in the schema definition above using AlsoRequires, ConflictsWith and the like. See https://developer.hashicorp.com/terraform/plugin/framework/migrating/attributes-blocks/validators-predefined

EvandroLG: I trust your Terraform expertise far more than my own, but these validations check specific field values (e.g. `type == "s3"`) across nested attributes, not just presence. Predefined validators like `AlsoRequires` seems to work best for simple "field A needs field B" cases (correct if I'm wrong, please). IMO, this business logic is clearer and more maintainable in `ModifyPlan` with explicit error messages.

Does that make sense to you? :)

whites11 (approved):

nunoadrego (approved):

PR #391, opened 2025-10-07
Remove concurrency from terraform provider for clickpipes 

tpanetti:
# Why
We are deprecating the usage of Object Storage concurrency scaling from the scaling endpoint in favor of using the advanced settings endpoint for Object Storage.
# What
This simply removes the concurrency member from the scaling request struct which was used to include it in the request.


kitop (approved):

whites11 (approved):
not used anywhere?

jkaflik (commented):
IMO, it makes sense to deprecate it first and then remove it when we release ClickPipes resources in stable builds.

Let's follow a deprecation procedure proposed by Hashicorp:
https://developer.hashicorp.com/terraform/plugin/framework/deprecations#provider-attribute-removal

PR #403, opened 2025-11-11
Add postgres CDC terraform provider for ClickPipes

tpanetti:
# Why
We want to add support in the terraform provider for CDC Postgres
# What 
This adds the models, resource, and API calls for configuring a CDC postgres clickpipe.

### DIFF
--- a/pkg/resource/clickpipe.go
+++ b/pkg/resource/clickpipe.go
@@ -774,9 +774,6 @@ func (c *ClickPipeResource) Schema(_ context.Context, _ resource.SchemaRequest, 
 					},
 				},
 				Required: true,
-				PlanModifiers: []planmodifier.Object{
-					objectplanmodifier.RequiresReplace(),
-				},
 			},
 			"destination": schema.SingleNestedAttribute{
 				Description: "The destination for the ClickPipe.",
@@ -1073,6 +1070,20 @@ func (c *ClickPipeResource) ModifyPlan(ctx context.Context, request resource.Mod
 				"ClickPipe in Internal Error state",
 				"ClickPipe is in an internal error state. Contact ClickHouse Cloud support for assistance. Your changes will be applied.",
 			)
+		}
+	}
+
+	// For Postgres sources, managed_table should always be false (tables managed via table_mappings)
+	// Override the default value to prevent inconsistency errors
+	var sourceModel models.ClickPipeSourceModel
+	if diags := plan.Source.As(ctx, &sourceModel, basetypes.ObjectAsOptions{}); !diags.HasError() {
+		if !sourceModel.Postgres.IsNull() {
+			var destinationModel models.ClickPipeDestinationModel
+			if diags := plan.Destination.As(ctx, &destinationModel, basetypes.ObjectAsOptions{}); !diags.HasError() {
+				destinationModel.ManagedTable = types.BoolValue(false)
+				plan.Destination = destinationModel.ObjectValue()
+				response.Diagnostics.Append(response.Plan.Set(ctx, plan)...)
+			}
 		}
 	}
 
@@ -1730,8 +1741,7 @@ func (c *ClickPipeResource) getStateCheckFunc(ctx context.Context, plan models.C
 	// For other pipes, wait for Running state or terminal error states
 	return func(state string) bool {
 		return state == api.ClickPipeRunningState ||
-			state == api.ClickPipeFailedState ||
-			state == api.ClickPipeInternalErrorState
+			state == api.ClickPipeFailedState
 	}
 }
 
@@ -2142,7 +2152,7 @@ func (c *ClickPipeResource) syncClickPipeState(ctx context.Context, state *model
 	// But managed_table should be preserved from state since user can configure it
 	if isPostgresPipe {
 		destinationModel.Table = types.StringNull()
-		destinationModel.ManagedTable = stateDestinationModel.ManagedTable // Preserve from state
+		destinationModel.ManagedTable = types.BoolValue(false) // Always false for Postgres pipes
 		destinationModel.Columns = types.ListNull(models.ClickPipeDestinationColumnModel{}.ObjectType())
 		destinationModel.TableDefinition = types.ObjectNull(models.ClickPipeDestinationTableDefinitionModel{}.ObjectType().AttrTypes)
 	} else {


pkg/resource/clickpipe.go
```diff
@@ -567,6 +568,189 @@ func (c *ClickPipeResource) Schema(_ context.Context, _ resource.SchemaRequest,
...
+										},
+										PlanModifiers: []planmodifier.Int64{
+											int64planmodifier.RequiresReplace(),
+										},
+									},
+									"enable_failover_slots": schema.BoolAttribute{
+										Description: "Enable failover for created replication slot. Requires a replication slot to NOT be set.",
+										Optional:    true,
+										PlanModifiers: []planmodifier.Bool{
+											boolplanmodifier.RequiresReplace(),
```
whites11: do we really need to replace the pipe for any setting change?

pkg/resource/clickpipe.go
```diff
@@ -896,69 +1083,99 @@ func (c *ClickPipeResource) Create(ctx context.Context, request resource.CreateR
...
+				"destination.table is required for non-Postgres CDC sources.",
+			)
+			return
+		}
+
+		if destinationModel.Columns.IsNull() || len(destinationModel.Columns.Elements()) == 0 {
 			response.Diagnostics.AddError(
 				"Error Creating ClickPipe",
-				"Managed table requires table definition",
+				"destination.columns is required for non-Postgres CDC sources.",
```
whites11: I'd rather put a positive rather than negated error message.
`destination.columns is required for 's3' source.`

examples/resources/clickhouse_clickpipe/resource_postgres_cdc.tf
```diff
@@ -0,0 +1,82 @@
+resource "clickhouse_clickpipe" "postgres_cdc_clickpipe" {
+  name        = "My Postgres CDC ClickPipe"
+  description = "Change Data Capture pipeline from PostgreSQL to ClickHouse"
+
+  service_id = "e9465b4b-f7e5-4937-8e21-8d508b02843d"
+
+  state = "Running"
```
whites11: this is wrong isn't it

whites11 (approved):
LGTM

pkg/resource/clickpipe.go
```diff
@@ -567,6 +568,189 @@ func (c *ClickPipeResource) Schema(_ context.Context, _ resource.SchemaRequest,
...
 					},
+					"postgres": schema.SingleNestedAttribute{
+						MarkdownDescription: "The Postgres CDC source configuration for the ClickPipe.",
+						Optional:            true,
+						Attributes: map[string]schema.Attribute{
+							"host": schema.StringAttribute{
+								Description: "The hostname of the Postgres instance.",
+								Required:    true,
+								PlanModifiers: []planmodifier.String{
+									stringplanmodifier.RequiresReplace(),
```
ilidemi: OpenAPI allows host, port and database to be updated in-place. Host, for example, is useful when switching from direct connection to private link

pkg/resource/clickpipe.go
```diff
@@ -1070,6 +1288,19 @@ func (c *ClickPipeResource) Create(ctx context.Context, request resource.CreateR
 	response.Diagnostics.Append(diags...)
 }
 
+func getSourceType(sourceModel models.ClickPipeSourceModel) string {
```
ilidemi: Should this return an enum?

pkg/resource/clickpipe.go
```diff
@@ -1520,43 +1849,235 @@ func (c *ClickPipeResource) syncClickPipeState(ctx context.Context, state *model
...
+		}
+
+		// For optional string fields, if state was null, keep it null (API may return empty string)
+		if stateSettingsModel.PublicationName.IsNull() {
+			settingsModel.PublicationName = types.StringNull()
+		} else if clickPipe.Source.Postgres.Settings.PublicationName != nil && *clickPipe.Source.Postgres.Settings.PublicationName != "" {
+			settingsModel.PublicationName = types.StringValue(*clickPipe.Source.Postgres.Settings.PublicationName)
+		} else {
+			settingsModel.PublicationName = stateSettingsModel.PublicationName
+		}
```
ilidemi: Trying to wrap my head around this. If we're syncing the state from the API, wouldn't this be:
```go
if clickPipe.Source.Postgres.Settings.PublicationName != nil &&
   *clickPipe.Source.Postgres.Settings.PublicationName != "" {
    settingsModel.PublicationName = types.StringValue(*clickPipe.Source.Postgres.Settings.PublicationName)
} else {
    settingsModel.PublicationName = types.StringNull()
}
```
We're not _expecting_ desyncs but if the local state got corrupted and wasn't null but became null, isn't the sync's job to bring it back from null

pkg/resource/clickpipe.go
```diff
@@ -1520,43 +1849,235 @@ func (c *ClickPipeResource) syncClickPipeState(ctx context.Context, state *model
...
+	destinationModel := models.ClickPipeDestinationModel{
+		Database: types.StringValue(clickPipe.Destination.Database),
+	}
+
+	// For Postgres CDC, the API doesn't return table/managedTable/columns, so preserve state values
+	if isPostgresPipe {
+		destinationModel.Table = stateDestinationModel.Table
+		destinationModel.ManagedTable = stateDestinationModel.ManagedTable
+		destinationModel.Columns = stateDestinationModel.Columns
+		destinationModel.TableDefinition = stateDestinationModel.TableDefinition
```
ilidemi: Should those be nulled out as they're not applicable?

pkg/resource/clickpipe.go
ilidemi: If this makes sense, should we explicitly map just (API default, state null) to (new state null) and not (API any, state null)?

examples/resources/clickhouse_clickpipe/resource_postgres_cdc.tf
```diff
@@ -0,0 +1,82 @@
...
+        target_table       = "users"
+
+        # Optional: Columns to exclude from replication
+        # excluded_columns = ["password_hash", "internal_notes"]
+
+        # Optional: Use custom sorting key
+        # use_custom_sorting_key = true
+        # sorting_keys = ["id", "created_at"]
+
+        # Optional: Specify table engine (default: ReplacingMergeTree for CDC)
```
ilidemi:
```suggestion
        # Optional: Specify table engine (default: ReplacingMergeTree)
```
(I think it's default everywhere in postgres?)

ilidemi (commented):
Thanks for taking this 🙏

pkg/resource/clickpipe.go
ilidemi: Modifying this setting is supported in PeerDB but not exposed to UI/OpenAPI yet, left comments on the ones that don't need replace. Claude is saying that moving from replace to no plan modifiers is safe with a version bump so I think we're good here.

Newbie questions on replace:
1) If in our case replace means resync, and for some customers that can be hours/days with terabytes of traffic and data unavailable in between, what is the norm around signaling how destructive the change is in TF? Could be a measured decision or not paying attention when modifying an integer
2) Replace would require deleting the destination tables, DB pipes don't do that automatically to avoid surprises and block the creation if the table is not empty. How can this be modeled/implemented?
3) cdc_only pipes are ok with the destination table not being empty but if the replication slot (the thing that remembers how far you've read and tells PG to keep the WAL) was created by us it will be deleted and miss updates during recreation. Replacing a cdc_only pipe with customer's replication slot is safe however. 
3) If a destination table is a MergeTree, deleting that would be a destructive operation as source Postgres wouldn't have these changes anymore, what do?

pkg/resource/clickpipe.go
```diff
@@ -567,6 +568,189 @@ func (c *ClickPipeResource) Schema(_ context.Context, _ resource.SchemaRequest,
...
+												stringvalidator.OneOf(api.ClickPipePostgresTableEngines...),
+											},
+										},
+									},
+								},
+							},
+						},
+						PlanModifiers: []planmodifier.Object{
+							objectplanmodifier.RequiresReplace(),
+						},
```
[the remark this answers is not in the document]
tpanetti: Ah yes, didn't realize I had this on the entire postgres object, fixing.

Also adding the pipeUpdate block and fixing the error message.

pkg/resource/clickpipe.go
tpanetti: Yeah I think I had this backwards, switched the mapping now.

pkg/resource/clickpipe.go
```diff
@@ -1275,13 +1291,43 @@ func (c *ClickPipeResource) Create(ctx context.Context, request resource.CreateR
...
+						}
+					}
+				}
+			}
+		}
+
+		// For snapshot-only pipes, accept both Running and Completed states
+		if isSnapshotOnly {
+			stateCheckFunc = func(state string) bool {
+				return state == api.ClickPipeRunningState || state == api.ClickPipeCompletedState
```
ilidemi: Running state means CDC, snapshot would go from `Snapshot` to `Completed` (or `Terminating`/`Terminated`/`Failed`)

pkg/resource/clickpipe.go
```diff
@@ -1937,13 +1978,16 @@ func (c *ClickPipeResource) syncClickPipeState(ctx context.Context, state *model
...
 
-		// For optional bool fields, if state was null, keep it null
-		if stateSettingsModel.EnableFailoverSlots.IsNull() {
-			settingsModel.EnableFailoverSlots = types.BoolNull()
-		} else if clickPipe.Source.Postgres.Settings.EnableFailoverSlots != nil {
-			settingsModel.EnableFailoverSlots = types.BoolValue(*clickPipe.Source.Postgres.Settings.EnableFailoverSlots)
+		// Preserve null from state if it was null and API returns a default
+		if clickPipe.Source.Postgres.Settings.EnableFailoverSlots != nil {
+			// If state had null, preserve it
+			if !stateSettingsModel.EnableFailoverSlots.IsNull() {
```
ilidemi: If state has null and clickpipe says true, state should become true

pkg/resource/clickpipe.go
ilidemi: `Snapshot` || `Completed` for snapshot-only pipes

pkg/resource/clickpipe.go
```diff
@@ -1275,13 +1291,43 @@ func (c *ClickPipeResource) Create(ctx context.Context, request resource.CreateR
...
+			stateCheckFunc = func(state string) bool {
+				return state == api.ClickPipeRunningState
+			}
+		}
 	}
-	if _, err := c.client.WaitForClickPipeState(ctx, serviceID, createdClickPipe.ID, func(state string) bool {
-		return state == expectedState
-	}, clickPipeStateChangeMaxWaitSeconds); err != nil {
+
+	if _, err := c.client.WaitForClickPipeState(ctx, serviceID, createdClickPipe.ID, stateCheckFunc, clickPipeStateChangeMaxWaitSeconds); err != nil {
```
ilidemi: What would be the golden path for CDC pipes with a 1h-1d snapshot?

pkg/internal/api/clickpipe_models.go
```diff
@@ -144,12 +144,14 @@ type ClickPipeKinesisSource struct {
...
-	Database    string                          `json:"database"`
-	Credentials *ClickPipeSourceCredentials     `json:"credentials,omitempty"`
-	Settings    ClickPipePostgresSettings       `json:"settings"`
-	Mappings    []ClickPipePostgresTableMapping `json:"tableMappings"`
+	Host                  string                          `json:"host"`
+	Port                  int                             `json:"port"`
+	Database              string                          `json:"database"`
+	Credentials           *ClickPipeSourceCredentials     `json:"credentials,omitempty"`
+	Settings              ClickPipePostgresSettings       `json:"settings"`
+	Mappings              []ClickPipePostgresTableMapping `json:"tableMappings"`
```
ilidemi: Should Mappings be `omitempty` too (if we're using same contracts for create and update calls)?

pkg/resource/clickpipe.go
ilidemi: To be more clear: for CDC pipes these are always null, so if state has a random value by mistake, it has to be overridden with null

ilidemi (commented):
Validating table mappings seems more complex than originally thought, I would move it to a separate PR after pause support and make sure it's rock solid there, and have them here as RequiresReplace. This would be important because changes there have to be applied in multiple API calls, and if we defer validation to the API, it can fail midway and leave the pipe in a state inconsistent with a plan.

Specific points:
* Mapping key would be (source schema, source table)
* Check the keys for conflicts and error out if any
* Target table can't conflict with any of the existing ones or newly added ones - will fail in API validation or at runtime otherwise
* User can try to change engine/target table/excluded columns/etc on an existing mapping, and that requires full removal then a full add with resync, with a clear warning during plan.
* Thorough tests for all of these

Also, to fully support resource replace we need to truncate destination table with SQL. Should it be in this PR or a separate one?

tpanetti:
> To be more clear: for CDC pipes these are always null, so if state has a random value by mistake, it has to be overridden with null
Looks like this is true for all of those variables except managed_table which is returned by the API. Fixed.

pkg/resource/clickpipe.go
```diff
@@ -1534,43 +1961,229 @@ func (c *ClickPipeResource) syncClickPipeState(ctx context.Context, state *model
...
 
+	destinationModel := models.ClickPipeDestinationModel{
+		Database: types.StringValue(clickPipe.Destination.Database),
+	}
+
+	// For Postgres CDC, table/columns/tableDefinition are always null (managed via table_mappings)
+	// But managed_table should be preserved from state since user can configure it
+	if isPostgresPipe {
+		destinationModel.Table = types.StringNull()
+		destinationModel.ManagedTable = stateDestinationModel.ManagedTable // Preserve from state
```
ilidemi: ManagedTable property doesn't apply to CDC pipes - it would be always null, if the contract allows, or always false. If state has anything else, it's the job of this function to reconcile

ilidemi (approved):
LGTM for this PR, just a few nits.

To bring follow-up threads in one place:
- It may take hours to days for a pipe to come to a RUNNING state, this should be allowed
- Get RequireReplace to work, as destination tables need to be deleted/truncated for that
- Pausing
- [Table mappings](https://github.com/ClickHouse/terraform-provider-clickhouse/pull/403#pullrequestreview-3502189255) changes/validations/planning
- Scaling
- Updates after OpenAPI-UI parity
- Source field changes (postgres->kafka) are now allowed in code, ideally need a require replace decision from code or to be rejected

pkg/resource/clickpipe.go
tpanetti: Moving to follow up PR.

PR #415, opened 2025-12-04
Fix Postgres CDC issues (table mappings, error handling, state, CDC scaling, etc)

tpanetti:
This PR fixes multiple issues with the Postgres CDC provider including
* Allow editable table mappings in a single API call
* Convert table mappings to a set to avoid order issues in the plan
* Add better error handling for table mappings (no single mapping pipe changes, no edits to the same source table in one API call).
* Fix source validation to disallow source type changes (general improvement)
* Add warning for deleting pipes to warn that the user needs to manually delete their destination tables
* Allow snapshot state as an acceptable state so both snapshot and CDC replication modes will provision faster and successfully on large loads
* Creates a new resource for CDC scaling which calls the endpoint for raising the scaling of CDC pipes (globally across a service).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/hashicorp/terraform-plugin-docs v0.24.0
 	github.com/hashicorp/terraform-plugin-framework v1.16.1
 	github.com/hashicorp/terraform-plugin-framework-validators v0.19.0
+	github.com/hashicorp/terraform-plugin-go v0.29.0
 	github.com/hashicorp/terraform-plugin-log v0.10.0
 	github.com/stretchr/testify v1.10.0
 	k8s.io/apimachinery v0.34.1
@@ -45,7 +46,6 @@ require (
 	github.com/hashicorp/hc-install v0.9.2 // indirect
 	github.com/hashicorp/terraform-exec v0.24.0 // indirect
 	github.com/hashicorp/terraform-json v0.27.2 // indirect
-	github.com/hashicorp/terraform-plugin-go v0.29.0 // indirect
 	github.com/hashicorp/terraform-registry-address v0.4.0 // indirect
 	github.com/hashicorp/terraform-svchost v0.1.1 // indirect
 	github.com/hashicorp/yamux v0.1.2 // indirect


pkg/resource/clickpipe_cdc_infrastructure.go
```diff
@@ -0,0 +1,275 @@
...
+		resp.Diagnostics.AddError(
+			"Error Creating CDC Infrastructure",
+			fmt.Sprintf("Could not create CDC infrastructure: %s", err.Error()),
+		)
+		return
+	}
+
+	// Update plan with response values
+	plan.ReplicaCpuMillicores = types.Int64Value(scaling.ReplicaCpuMillicores)
+	plan.ReplicaMemoryGb = types.Float64Value(scaling.ReplicaMemoryGb)
```
ilidemi: After UpdateCDCScaling returns, the values are "accepted", not "applied" - to properly update the plan we should wait till GetCDCScaling starts returning them (average 2-5 min)

pkg/resource/clickpipe.go
```diff
@@ -2047,27 +2146,32 @@ func (c *ClickPipeResource) syncClickPipeState(ctx context.Context, state *model
 			settingsModel.EnableFailoverSlots = types.BoolNull()
 		}
 
-		// Table mappings - preserve null values from state
-		var stateTableMappings []models.ClickPipePostgresTableMappingModel
+		// Table mappings - convert API response to Set (order doesn't matter)
+		// Get state mappings for preserving null values on optional fields
+		var stateTableMappingsMap map[string]models.ClickPipePostgresTableMappingModel
```
ilidemi: The key here should be `struct{sourceSchemaName, sourceTable string}` (simple structs can be keys in Go) - each source table can only be present once, and I'm not sure if it's completely impossible to have dots in names

pkg/resource/clickpipe.go
```diff
@@ -1094,6 +1100,83 @@ func (c *ClickPipeResource) ModifyPlan(ctx context.Context, request resource.Mod
...
+								stateMappings = make([]models.ClickPipePostgresTableMappingModel, len(statePostgres.TableMappings.Elements()))
+								statePostgres.TableMappings.ElementsAs(ctx, &stateMappings, false)
+							}
+							if !planPostgres.TableMappings.IsNull() {
+								planMappings = make([]models.ClickPipePostgresTableMappingModel, len(planPostgres.TableMappings.Elements()))
+								planPostgres.TableMappings.ElementsAs(ctx, &planMappings, false)
+							}
+
+							// Validation 1: Cannot delete the last table mapping
+							if len(stateMappings) == 1 && len(planMappings) == 0 {
```
ilidemi: Are stateMappings guaranteed to go through 1 before attempting 0? Going from 2 to 0 would also be a mistake

ilidemi (commented):
State looks good
Scaling looks good, with minor comments (feel free to check in after applying)

Deletion - @whites11 would defer to you on what is the best experience for "this plan assumes you're ok with deleting your data, you shouldn't really do this". Regular deletion doesn't need a warning IMO, if that is at all separate from replace

Table mappings - this will need careful iteration, please split out in a different PR which we'll take after I'm back

pkg/resource/clickpipe_cdc_infrastructure.go
tpanetti: Makes sense, I've switched now to adding a polling function after creation and update that checks for the values. Oddly though it seems to happen much faster than you describe (circa ~15 seconds). Might be worth looking into.

pkg/resource/clickpipe.go
```diff
@@ -2637,6 +2745,17 @@ func (c *ClickPipeResource) Delete(ctx context.Context, request resource.DeleteR
 		return
 	}
 
+	// Check if this is a Postgres pipe - warn about manual table cleanup
+	var sourceModel models.ClickPipeSourceModel
+	if diags := state.Source.As(ctx, &sourceModel, basetypes.ObjectAsOptions{}); !diags.HasError() {
+		if !sourceModel.Postgres.IsNull() {
+			response.Diagnostics.AddWarning(
+				"Manual table cleanup required",
```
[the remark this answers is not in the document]
tpanetti: Changed to detect the replace on ModifyPlan and throw the warning there!

pkg/resource/clickpipe.go
tpanetti: Yes you are correct, I think my head was wrapped in the logic of deleting/adding new mappings on the same request. We should just check that planMappings is never 0 as this will cause OpenAPI validation to fail.

pkg/resource/clickpipe_cdc_infrastructure.go
```diff
@@ -0,0 +1,303 @@
...
+		Attributes: map[string]schema.Attribute{
+			"service_id": schema.StringAttribute{
+				MarkdownDescription: "ClickHouse Cloud service ID where the CDC infrastructure is located.",
+				Required:            true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"replica_cpu_millicores": schema.Int64Attribute{
+				MarkdownDescription: "CPU in millicores for DB ClickPipes. Must be a multiple of 1000, between 1000 and 24000.",
```
whites11: how do we validate it's a multiple of 1k? server side?

pkg/resource/clickpipe_cdc_infrastructure.go
```diff
@@ -0,0 +1,303 @@
...
+			fmt.Sprintf("replica_memory_gb must be 4× the CPU core count. With %d millicores (%.1f cores), memory must be %.1f GB, but got %.1f GB",
+				cpuMillicores, cpuCores, expectedMemory, memoryGb),
+		)
+		return
+	}
+
+	// Poll for CDC scaling endpoint to become available (up to 10 minutes)
+	maxWait := 10 * time.Minute
+	checkAvailable := func() error {
+		_, err := r.client.GetClickPipeCdcScaling(ctx, serviceID)
```
whites11: why do we need this at all rather than failing the apply when users tries to create a cdc_infrastructure but has no pipes?

whites11 (approved):

PR #431, opened 2026-01-07
feat: Add write only password attribute (`password_wo`) to Clickhouse service

adriancooney:
## Summary
Introduces a [write-only password attribute](https://developer.hashicorp.com/terraform/language/manage-sensitive-data/write-only) for ClickHouse services to avoid storing password hashes in Terraform state.

### Changes
- Added `password_wo` (write-only, not persisted to state)
- Added `password_wo_version` to trigger password updates when using `password_wo`
- Updated validators to ensure mutual exclusivity between `password`, `password_wo`, `password_hash`, `double_sha1_password_hash`, and `warehouse_id`

### Usage

```tf 
resource "clickhouse_service" "example" {
  name           = "my-service"
  cloud_provider = "aws"
  region         = "us-east-1"
  
  password_wo         = var.clickhouse_password
  password_wo_version = 1  # Increment to trigger password update
  
  ip_access = [...]
}
```

### Why
The existing `password` or `password_hash` attribute stores a plaintext/SHA256 hash in Terraform state, which is a security concern. The new `password_wo` attribute uses Terraform's write-only feature to accept the password during apply but never persist it to state.

### DIFF
--- a/docs/resources/service.md
+++ b/docs/resources/service.md
@@ -57,11 +57,13 @@ resource "clickhouse_service" "service" {
 
 ### Optional
 
+> **NOTE**: [Write-only arguments](https://developer.hashicorp.com/terraform/language/resources/ephemeral#write-only-arguments) are supported in Terraform 1.11 and later.
+
 - `backup_configuration` (Attributes) Configuration of service backup settings. (see [below for nested schema](#nestedatt--backup_configuration))
 - `backup_id` (String) ID of the backup to restore when creating new service. If specified, the service will be created as a restore operation
 - `byoc_id` (String) BYOC ID related to the cloud provider account you want to create this service into.
 - `compliance_type` (String) Compliance type of the service. Can be 'hipaa', 'pci'. Required for organizations that wish to deploy their services in the hipaa/pci compliant environment. NOTE: hipaa/pci compliance should be enabled for your ClickHouse organization before using this field.
-- `double_sha1_password_hash` (String, Sensitive) Double SHA1 hash of password for connecting with the MySQL protocol. Cannot be specified if `password` is specified.
+- `double_sha1_password_hash` (String, Sensitive) Double SHA1 hash of password for connecting with the MySQL protocol. Cannot be specified if `password` or `password_wo` is specified.
 - `encryption_assumed_role_identifier` (String) Custom role identifier ARN.
 - `encryption_key` (String) Custom encryption key ARN.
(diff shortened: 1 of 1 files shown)


pkg/resource/service.go
nunoadrego:
```suggestion
				Description: "SHA256 hash of password for the default user. One of either `password`, `password_wo`, or `password_hash` must be specified.",
```

pkg/resource/service.go
```diff
@@ -163,6 +165,27 @@ func (r *ServiceResource) Schema(_ context.Context, _ resource.SchemaRequest, re
 					stringvalidator.AlsoRequires(path.Expressions{path.MatchRoot("password_hash")}...),
 				},
 			},
+			"password_wo": schema.StringAttribute{
+				Description: "Password for the default user (write-only, not persisted to state). Use this instead of `password` to avoid storing the password hash in Terraform state.",
+				Optional:    true,
+				Sensitive:   true,
+				WriteOnly:   true,
+				Validators: []validator.String{
```
nunoadrego: What do you think of adding that it also requires password_wo_version?

pkg/resource/service.go
```diff
@@ -1102,10 +1125,12 @@ func (r *ServiceResource) Create(ctx context.Context, req resource.CreateRequest
 
 	// Password and backup settings are only set on parent instances for hydra services
 	if plan.DataWarehouseID.IsUnknown() || plan.DataWarehouseID.IsNull() {
-		// Update service password if provided explicitly
-		planPassword := plan.Password.ValueString()
-		if len(planPassword) > 0 {
-			_, err := r.client.UpdateServicePassword(ctx, s.Id, servicePasswordUpdateFromPlainPassword(planPassword))
+		// Update service password if provided explicitly (prefer password_wo over password)
+		if password, passwordWO := plan.Password.ValueString(), plan.PasswordWO.ValueString(); len(password) > 0 || len(passwordWO) > 0 {
```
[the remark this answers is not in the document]
adriancooney: Ah I totally missed that PR. Good call, I've updated the code to read from the config.

nunoadrego (approved):

nunoadrego:
@adriancooney can you run make docs? It's the only check that's failing so far.

nunoadrego:
Bypass rules due to:
- E2E tests can't be run in this PR, but were run here: https://github.com/ClickHouse/terraform-provider-clickhouse/actions/runs/20970464995
- We have an old Terraform version in the runners that results in a different docs generated

PR #476, opened 2026-03-10
Add mysql clickpipe provider implementation

tpanetti:
# Why
We've recently introduced OpenAPI support for MySQL ClickPipes and want to add support in our terraform provider

# What
This adds support in our terraform provider for MySQL ClickPipes along with validation.


pkg/internal/api/clickpipe.go
```diff
@@ -182,6 +182,53 @@ var ClickPipeBigQueryTableEngines = []string{
 	ClickPipeTableEngineNull,
 }
 
+const (
+	ClickPipeMySQLReplicationMechanismAuto    = "AUTO"
```
ilidemi: FYI we're removing AUTO and defaulting to GTID, would be quicker to reflect it here in this change

PR #492, opened 2026-03-19
Move ClickPipes provider to GA

tpanetti:
# Why
We want to launch the ClickPipes provider now that we have support for all major clickpipe types.
# What
This removes the alpha tag from ClickPipes resources to allow the provider to build it into the general terraform release. It also does the following to prepare for GA
* Updates examples and documentation to match GA resources -> This includes better MDs, terraform examples, sync state script, etc
* Renames all E2E test files, dropping the alpha tag from the filename so the script will run them with the GA provider.
* Updates the docs snapshot with the new ClickPipes docs now that they're included in GA.
* Minor fix to clickpipes_cdc_infra to allow import state which was missed

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -469,6 +469,18 @@ jobs:
           go-version-file: 'go.mod'
           cache: true
 
+      - name: Get previous stable tag
+        id: prev_tag
+        if: ${{ needs.validate.outputs.enable_alpha == 'false' }}
+        run: |
+          current="v${{ inputs.version }}"
+          # Find the most recent stable tag (no pre-release suffix) before the current one
+          prev=$(git tag --list 'v*' --sort=-version:refname | grep -v -- '-' | grep -v "^${current}$" | head -1)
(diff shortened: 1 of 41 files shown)


pkg/resource/descriptions/clickpipe.md
```diff
@@ -0,0 +1,8 @@
+You can use the *clickhouse_clickpipe* resource to create and manage ClickPipes data ingestion pipelines in ClickHouse Cloud.
+
+Supported source types: Kafka (Confluent, MSK, Azure Event Hubs, Redpanda, WarpStream), Object Storage (S3, GCS, Azure Blob), Kinesis, Postgres CDC, MySQL CDC, BigQuery, and MongoDB CDC.
+
+Known limitations:
+
+- ClickPipe does not support table updates for managed tables. If you need to update the table schema, you will have to do that externally.
```
jkaflik: Let's discuss that internally. This feature is wonky and I believe it's a good time to make a good decision around it before GA.

PR #510, opened 2026-04-16
ci(releases): move release notes steps to dedicated workflow and use langfuse

nunoadrego:
Extract AI release notes generation into a standalone workflow and add Langfuse observability to the Claude Code action using the hooks system.


.claude/hooks/langfuse_hook.py
```diff
@@ -0,0 +1,550 @@
+#!/usr/bin/env python3
```
nunoadrego, line 1: From: https://langfuse.com/integrations/other/claude-code#quick-start

fjgcosta (approved):

PR #512, opened 2026-04-17
feat: add GCP PSC Service Attachment support for RPEs

EvandroLG:
## Summary

- Add `GCP_PSC_SERVICE_ATTACHMENT` as a supported type for `clickhouse_clickpipes_reverse_private_endpoint`, with a new optional `gcp_service_attachment`
  attribute (format: `projects/{project}/regions/{region}/serviceAttachments/{name}`).
- Update the embedded resource description and generated docs to list the new type.
- Add a schema unit test to guard the attribute and the extended type enum against regressions.

### DIFF
--- a/pkg/resource/clickpipe_reverse_private_endpoint.go
+++ b/pkg/resource/clickpipe_reverse_private_endpoint.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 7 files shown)


pkg/resource/clickpipe_reverse_private_endpoint.go
```diff
@@ -196,6 +203,14 @@ func (r *ClickPipeReversePrivateEndpointResource) Create(ctx context.Context, re
 			)
 			return
 		}
+	case api.ReversePrivateEndpointTypeGcpPscServiceAttachment:
+		if data.GcpServiceAttachment.IsNull() {
+			resp.Diagnostics.AddError(
+				"Missing required field",
+				"gcp_service_attachment is required when type is GCP_PSC_SERVICE_ATTACHMENT",
+			)
+			return
```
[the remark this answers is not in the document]
EvandroLG: I don't think this is reachable here. `gcp_service_attachment` is Optional, so by the time `Create` runs the framework guarantees it's either `Null` or a concrete string.

pkg/resource/clickpipe_reverse_private_endpoint.go
```diff
@@ -119,6 +119,13 @@ func (r *ClickPipeReversePrivateEndpointResource) Schema(ctx context.Context, re
 					stringplanmodifier.RequiresReplace(),
 				},
 			},
+			"gcp_service_attachment": schema.StringAttribute{
+				Optional:            true,
+				MarkdownDescription: "GCP PSC service attachment URI, required for GCP_PSC_SERVICE_ATTACHMENT type. Format: projects/{project}/regions/{region}/serviceAttachments/{name}",
```
whites11, line 124: should we validate the format as described in the description with a stringvalidator.RegexMatches ?

whites11 (approved):
one suggestion, otherwise looks good

pkg/resource/clickpipe_reverse_private_endpoint.go
EvandroLG: Good call! Thanks, @whites11

jkaflik (approved):
LGTM. As we discussed internally, let's hold with the merge until the release day.

PR #527, opened 2026-05-06
Add pubsub support for ClickPipes terraform resource

tpanetti:
# What
This adds support for Pub/Sub for ClickPipes terraform resource


pkg/resource/clickpipe.go
```diff
@@ -635,6 +637,126 @@ func (c *ClickPipeResource) Schema(_ context.Context, _ resource.SchemaRequest,
...
+										regexp.MustCompile(`^[a-z][a-z0-9-]{4,28}[a-z0-9]$`),
+										"must be a valid GCP project ID (6–30 chars, lowercase, digits and hyphens; cannot start with a digit or end with a hyphen)",
+									),
+								},
+								PlanModifiers: []planmodifier.String{
+									stringplanmodifier.RequiresReplace(),
+								},
+							},
+							"topic": schema.StringAttribute{
+								Description: "The Pub/Sub topic name (not the fully-qualified path).",
```
whites11, line 674: can we validate the format? "not the fully-qualified path"

pkg/resource/clickpipe.go
```diff
@@ -635,6 +637,126 @@ func (c *ClickPipeResource) Schema(_ context.Context, _ resource.SchemaRequest,
...
+								Validators: []validator.String{
+									stringvalidator.OneOf(api.ClickPipePubSubSeekTypes...),
+								},
+								PlanModifiers: []planmodifier.String{
+									stringplanmodifier.RequiresReplace(),
+								},
+							},
+							"seek_timestamp": schema.StringAttribute{
+								MarkdownDescription: fmt.Sprintf(
+									"RFC 3339 timestamp (e.g. `2026-04-10T12:00:00Z`). Required when `seek_type = \"%s\"`; must be omitted otherwise.",
```
whites11, line 711: I think you can validate the requiredness of a field based on the presence of another

whites11 (approved):

pkg/resource/clickpipe.go
tpanetti: The cross-field requiredness is enforced by `pubsubSeekValidator` in `pkg/resource/clickpipe_config_validators.go:18-114`, registered via `ConfigValidators()` at line 116. It produces plan-time errors like `seek_timestamp is required when seek_type is "timestamp"` and `seek_timestamp must not be set when seek_type is "earliest"` covering all four seek-type/companion combinations.

The plugin-framework doesn't have a built-in attribute-level validator for "required if another attribute equals X" (only `AlsoRequires` / `ConflictsWith` based on presence). The `ConfigValidator` pattern is the canonical way to express value-conditional requiredness in this framework. Happy to add a cross-reference comment on the schema attributes pointing at the validator if that helps discoverability.

tpanetti: Yes! We do this in the clickpipe_config_validators.go file.

whites11 (approved):

nunoadrego (approved):

PR #536, opened 2026-05-13
feat: add clickhouse_service_scheduled_scaling resource

serdec:
## Summary

Adds a new top-level resource `clickhouse_service_scheduled_scaling` that manages time-based auto-scaling schedules for ClickHouse Cloud services. The resource consumes the public `/v1/organizations/{orgId}/services/{instanceId}/scalingSchedule` API (GET/POST/DELETE).

A schedule is a list of recurring weekly windows. While a window is active the service uses the entry's replica count, memory bounds, and idle-scaling settings; otherwise it falls back to base auto-scaling.

```hcl
resource "clickhouse_service_scheduled_scaling" "example" {
  service_id = clickhouse_service.example.id

  entries = [
    {
      name           = "Business hours"
      weekdays       = [1, 2, 3, 4, 5]
      start_hour_utc = 8
      end_hour_utc   = 18
      min_replicas   = 3
      max_replicas   = 3
      idle_scaling   = false
    },
    {
      name                 = "Overnight"
      weekdays             = [0, 1, 2, 3, 4, 5, 6]
      start_hour_utc       = 22
      end_hour_utc         = 6 # wraps overnight when end < start
      min_replicas         = 1
      max_replicas         = 1
      idle_scaling         = true
      idle_timeout_minutes = 5
    },
  ]
}
```

## Notes

- The feature is gated server-side by `canUseScheduledAutoscaling` (`FT_ORG_AUTOSCALING_SCHEDULE`); the API returns 403 when disabled, which surfaces to the user via the resource error.
- Server constraint: `min_replicas == max_replicas` per entry. Enforced client-side in `ValidateConfig` with a clearer error than the server's.
- Max 10 entries per service.
- Weekdays: 0 = Sunday, 6 = Saturday.
- `end_hour_utc` may be less than `start_hour_utc` to express an overnight window.
- Entry order on the server is preserved (MongoDB embedded array; confirmed via the OpenAPI handler and storage path).
- The server-side `is_active_now` / `active_entry_id` time-dependent fields are intentionally NOT exposed in TF state to avoid perpetual-drift loops on refresh.

## Test plan

- [x] `go build ./...` and `go build -tags alpha ./...`
- [x] `go vet ./...` and `go vet -tags alpha ./...`
- [x] `go test ./...` — all pass
- [x] Unit tests for API client (GET/POST/DELETE) via httptest
- [x] Unit tests for `applyScheduleToState`, `planEntriesToAPI`, `validateScheduledScalingEntries`
- [ ] Manual `terraform apply` against a dev org with `canUseScheduledAutoscaling` enabled


pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,475 @@
...
+							Optional:    true,
+						},
+						"idle_scaling": schema.BoolAttribute{
+							Description: "Whether idle scaling is enabled while the window is active.",
+							Optional:    true,
+						},
+						"idle_timeout_minutes": schema.Int64Attribute{
+							Description: "Minutes of inactivity before the service scales to zero (only relevant when idle_scaling is true).",
+							Optional:    true,
+						},
```
serdec, line 171: **Probable perpetual drift.** All six optional entry attributes are `Optional` only. Compare to `service.go:328-362` (`min_replica_memory_gb`, `max_replica_memory_gb`, `num_replicas`, `idle_timeout_minutes`) which are all `Optional + Computed` with `int64planmodifier.UseStateForUnknown()` / `boolplanmodifier.UseStateForUnknown()`.

If the server echoes back a value the caller didn't send (e.g. fills `idle_timeout_minutes` with a default when `idle_scaling=true`, or normalizes memory bounds), `Read` will write that value to state while the planned value stays `null` → diff on every plan → next `Update` tries to clear it → server may refuse → loop.

The PR description acknowledges this risk for `is_active_now`/`active_entry_id` but doesn't carry it through to defaulted fields. Suggest either matching the `service.go` pattern, or confirming with the OpenAPI handler that the server never returns fields the caller didn't send.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,475 @@
...
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"entries": schema.ListNestedAttribute{
+				Description: "Ordered list of recurring scaling windows. The first entry whose weekday and hour range covers \"now\" is applied; otherwise base_config applies.",
+				Required:    true,
+				Validators: []validator.List{
+					listvalidator.SizeAtMost(maxScheduledScalingEntries),
+				},
```
serdec: `Required` + `SizeAtMost(10)` but no `SizeAtLeast(1)`. A user can write `entries = []` and the resource will POST `{"entries":[]}`. Two failure modes depending on server semantics:

- If the server treats empty as "delete schedule", the next GET returns 404 → state removed → next plan recreates → infinite redo.
- If the server treats empty as a no-op, the resource just sits there meaninglessly (use `terraform destroy` to remove).

Either way, an empty list is never a useful user intent. Add `listvalidator.SizeAtLeast(1)`. Also makes the null-check in `planEntriesToAPI` (line 333) dead code, which is good — it currently masks the intent.

The API test `TestUpdateScheduledScaling_EmptyEntriesSerializesToEmptyArray` only proves the client serializes correctly — it doesn't validate the *server's* interpretation.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,475 @@
...
+	resp.Schema = schema.Schema{
+		MarkdownDescription: serviceScheduledScalingResourceDescription,
+		Attributes: map[string]schema.Attribute{
+			"service_id": schema.StringAttribute{
+				Description: "ClickHouse Cloud service ID this schedule applies to.",
+				Required:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
```
serdec, line 67: **Missing top-level `id` attribute.** Every other top-level resource in this provider exposes one — `role.go:51`, `service.go:70`, `organization_settings.go:49`, `clickpipe.go:123`, `clickpipe_reverse_private_endpoint.go:52`, `private_endpoint_registration.go`. Even single-tenant resources like `organization_settings` define it.

Impacts:
- `terraform state show` / list displays look odd without it.
- The framework's `resource.ImportStatePassthroughID` convention is to write to `path.Root("id")` (line 258 currently uses `service_id` instead).
- Some tooling assumes `id` exists.

Suggest mirroring `organization_settings.go:49-56`: add `id` as `Computed` + `UseStateForUnknown`, set it equal to `service_id` (or an `org/<orgId>/service/<serviceId>` composite) on Create/Read, import via `path.Root("id")`.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,475 @@
...
+				Required:    true,
+				Validators: []validator.List{
+					listvalidator.SizeAtMost(maxScheduledScalingEntries),
+				},
+				NestedObject: schema.NestedAttributeObject{
+					Attributes: map[string]schema.Attribute{
+						"id": schema.StringAttribute{
+							Description: "Server-generated entry ID. Shown as \"known after apply\" on any change to the entries list because IDs are assigned by the server in positional order.",
+							Computed:    true,
+						},
```
serdec: Without a plan modifier, this computed `id` will render as `(known after apply)` on every plan — even when the entries list is structurally unchanged — which is noisy enough that reviewers will assume something's actually changing.

Options:
- Drop the `id` field entirely. Nothing user-facing uses it AFAICT, and the PR description says position is the stable identity.
- Or add `stringplanmodifier.UseStateForUnknown()` (works only when the parent list is positionally stable — which the description claims).

Also: the description says "IDs are assigned by the server in positional order" — worth confirming: if a user inserts a new entry at position 1, do existing entries keep their IDs or do they shift? If they keep their IDs, the position-based assumption in the description is wrong.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,475 @@
...
+		return
+	}
+
+	entries, d := planEntriesToAPI(ctx, plan.Entries)
+	resp.Diagnostics.Append(d...)
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	schedule, err := r.client.UpdateScheduledScaling(ctx, plan.ServiceID.ValueString(), api.AutoScalingScheduleUpdate{Entries: entries})
```
serdec: `Create` calls `UpdateScheduledScaling` (POST) which is an idempotent overwrite. If a schedule already exists on the service (created via UI, leftover from a failed `terraform destroy`, etc.), Terraform will silently clobber it.

Convention is to GET first; if a schedule exists, error with "already exists, please import". Not a blocker, but a real footgun. See `clickpipe.go` Create for an example of pre-create checks.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,475 @@
...
+	if err != nil {
+		resp.Diagnostics.AddError("Error updating scheduled scaling", err.Error())
+		return
+	}
```
serdec, line 313: Inconsistent 404 handling. `Read` (line 201) and `Delete` (line 252) both treat `IsNotFound` specially. `Create` and `Update` don't. If the underlying service is deleted between plan and apply, the user gets a raw `status: 404` error instead of a friendlier message or a state-remove. Worth a thought; not a blocker.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,475 @@
...
+		if minRepSet && maxRepSet && e.MinReplicas.ValueInt64() != e.MaxReplicas.ValueInt64() {
+			diags.AddAttributeError(
+				entryPath,
+				"min_replicas must equal max_replicas",
+				"The scheduled scaling API currently requires a fixed replica count per entry.",
+			)
+		}
+	}
+	return diags
+}
```
serdec, line 405: Validation gaps relative to `service.go`:

- `idle_timeout_minutes` should require `idle_scaling=true` — `service.go:359` does this with `int64validator.AlsoRequires(path.Expressions{path.MatchRoot("idle_scaling")}...)`. Server presumably ignores it when `idle_scaling=false`, but the client-side rejection is friendlier.
- No minimum on `idle_timeout_minutes` (`service.go` requires `>= 5`).
- No check that `min_replica_memory_gb <= max_replica_memory_gb`.
- No positivity check on memory values.

None of these are correctness bugs (the server will reject), but they replicate one of the explicit benefits the PR description claims for client-side validation ("a clearer error than the server's").

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,475 @@
...
+	for i, em := range entryModels {
+		var weekdays []int64
+		diags.Append(em.Weekdays.ElementsAs(ctx, &weekdays, false)...)
+		if diags.HasError() {
+			return nil, diags
+		}
+		intWeekdays := make([]int, len(weekdays))
+		for j, w := range weekdays {
+			intWeekdays[j] = int(w)
+		}
```
serdec, line 431: **Non-deterministic JSON ordering.** `Weekdays` is a `Set`, and `ElementsAs` on a Set doesn't guarantee iteration order is stable across plan/apply invocations. The HTTP body sent to the server could be `[1,3,2,4,5]` one apply and `[2,1,3,5,4]` the next.

Round-trip into a Set on Read normalizes the diff away, so this isn't a correctness bug — but it makes debugging painful (wire-level diffs vary apply-to-apply) and may interact weirdly with any server-side change detection that compares JSON bytes.

Cheap fix: `sort.Ints(intWeekdays)` before assigning.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,475 @@
...
+						},
+						"start_hour_utc": schema.Int64Attribute{
+							Description: "Start hour in UTC (0-23). If end_hour_utc < start_hour_utc the window wraps overnight.",
+							Required:    true,
+							Validators: []validator.Int64{
+								int64validator.Between(0, 23),
+							},
+						},
+						"end_hour_utc": schema.Int64Attribute{
+							Description: "End hour in UTC (1-24). Must differ from start_hour_utc.",
```
serdec: Asymmetry worth calling out more loudly in the description: `start_hour_utc` is 0–23, `end_hour_utc` is 1–24. Someone will write `end_hour_utc = 0` thinking "midnight" and hit the validator. The combination `start=0, end=24` represents a 24-hour window which is fine, but worth an explicit example in the resource markdown.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,475 @@
...
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+
+	"github.com/ClickHouse/terraform-provider-clickhouse/pkg/internal/api"
+	"github.com/ClickHouse/terraform-provider-clickhouse/pkg/resource/models"
+)
+
+const maxScheduledScalingEntries = 10
```
serdec: Nit: hard-coded `10`. If the server raises this limit you need a code change in the provider before users can use it. Consider sourcing from a constant exported from `pkg/internal/api` so client and server stay in sync.

pkg/resource/service_scheduled_scaling_test.go
```diff
@@ -0,0 +1,296 @@
...
+	if len(got) != 1 {
+		t.Fatalf("len(got) = %d; want 1", len(got))
+	}
+	g := got[0]
+	if g.Name != "primary" || g.StartHourUtc != 9 || g.EndHourUtc != 17 {
+		t.Errorf("scalar fields mismatch: %+v", g)
+	}
+	if len(g.Weekdays) != 2 {
+		t.Errorf("weekdays len = %d; want 2", len(g.Weekdays))
+	}
```
serdec: Asserts length only — a bug that mangled the weekday values would pass. Assert the actual contents (sorted, since Set ordering varies):

```go
var gotDays []int
for _, d := range g.Weekdays { gotDays = append(gotDays, d) }
sort.Ints(gotDays)
if !reflect.DeepEqual(gotDays, []int{1, 3}) { ... }
```

serdec (commented):
Pass through with thoroughness focus — most comments inline. Top three to verify before merge:

1. **Perpetual drift risk on optional entry fields** — they're `Optional` only, where `service.go` uses `Optional+Computed+UseStateForUnknown` for the same conceptual fields. If the server echoes back any defaulted/normalized value, every `terraform plan` will diff. Strongly recommend running a manual apply followed by a no-op plan to confirm.
2. **Empty `entries = []` is allowed** — `Required` + `SizeAtMost(10)` but no `SizeAtLeast(1)`. Combined with the POST being idempotent overwrite, this and a 404-on-GET-empty server behavior could put refresh/plan into a recreate loop.
3. **Missing top-level `id`** — every other resource in this provider exposes one (see `role.go:51`, `service.go:70`, `organization_settings.go:49`). Convention violation; some tooling assumes it.

Nothing here is a blocker on its own; #1 is the one I'd not ship without verifying.

pkg/resource/service_scheduled_scaling_test.go
```diff
@@ -0,0 +1,296 @@
...
+	if g.MinReplicaMemoryGb != nil || g.MaxReplicaMemoryGb != nil {
+		t.Errorf("memory pointers should be nil, got %v / %v", g.MinReplicaMemoryGb, g.MaxReplicaMemoryGb)
+	}
+	if g.MinReplicas != nil || g.MaxReplicas != nil {
+		t.Errorf("replica pointers should be nil, got %v / %v", g.MinReplicas, g.MaxReplicas)
+	}
+	if g.IdleScaling != nil || g.IdleTimeoutMinutes != nil {
+		t.Errorf("idle pointers should be nil, got %v / %v", g.IdleScaling, g.IdleTimeoutMinutes)
+	}
+}
```
serdec, line 389: Coverage gaps worth filling:

- **Multi-entry test** — every `TestApplyScheduleToState_*` only exercises a single entry. The order-preservation guarantee in the PR description is untested.
- **Round-trip stability** — apply → state → plan-no-op test. This would catch the `Optional`-vs-`Optional+Computed` drift issue (see the comment on lines 103-126) without needing a live server.
- **Server-normalizes-values scenario** — explicit test where the API response includes fields the user didn't set, asserting that state matches what the user intended (or, if you keep the current schema, that state matches the server response and you've accepted the trade-off).
- **ImportState** — no test exists.
- The unchecked checkbox in the PR description ("Manual `terraform apply` against a dev org") is the single most valuable test left given the drift risk.

pkg/internal/api/scheduled_scaling_test.go
```diff
@@ -0,0 +1,164 @@
...
+	})
+
+	_, err := client.UpdateScheduledScaling(context.Background(), "svc-1", AutoScalingScheduleUpdate{Entries: []AutoScalingScheduleEntry{}})
+	if err != nil {
+		t.Fatalf("UpdateScheduledScaling: %v", err)
+	}
+	if !strings.Contains(captured, `"entries":[]`) {
+		t.Errorf("request body should contain empty entries array, got: %s", captured)
+	}
+}
```
serdec, line 147: This test only proves client-side serialization. It doesn't tell you what the *server* does with `{"entries":[]}` — which is the ambiguity that matters (delete schedule vs. no-op). See the comment on the resource's `entries` schema definition; recommend adding `SizeAtLeast(1)` so this code path is never exercised in practice.

serdec:
Thanks for the thorough review. Pushed [410adc7](https://github.com/ClickHouse/terraform-provider-clickhouse/pull/536/commits/410adc7) addressing every inline comment.

### Top three concerns
1. **Perpetual drift on optional entry fields** — fixed. Every optional entry attribute (`min_replica_memory_gb`, `max_replica_memory_gb`, `min_replicas`, `max_replicas`, `idle_scaling`, `idle_timeout_minutes`) is now `Optional + Computed + UseStateForUnknown`, matching the `service.go` pattern. If the server fills a defaulted/normalized value, state captures it and subsequent plans treat it as the source of truth.
2. **Empty `entries = []` allowed** — fixed. Added `listvalidator.SizeAtLeast(1)`. The configuration is rejected before reaching the API. The corresponding null/empty branch in `planEntriesToAPI` is no longer reachable from a real plan but kept as a defense-in-depth client-only guard.
3. **Missing top-level `id`** — fixed. Added a `Computed` `id` (= `service_id`) with `UseStateForUnknown`, and switched `ImportState` to write both `id` and `service_id` so the import path matches the convention used by `role.go` / `organization_settings.go`.

### Inline comments
| # | Comment | Resolution |
|---|---------|------------|
| 1 | Drift on Optional entry fields | Made all six fields `Optional+Computed+UseStateForUnknown` (matches `service.go:328-362`) |
| 2 | `SizeAtLeast(1)` on entries | Added |
| 3 | Top-level `id` attribute | Added; `ImportState` writes both `id` and `service_id` |
| 4 | Entry-level `id` shows "known after apply" every plan | Dropped the entry-level `id` field — not user-facing; eliminates the noise altogether |
| 5 | Create silently clobbers existing schedule | `Create` now `GET`s first; refuses with a "please import" error if a schedule with entries already exists |
| 6 | Inconsistent 404 handling | `Create` and `Update` now produce a friendly "Service not found" error on 404 instead of the raw status string |
| 7 | Validation gaps | Added `int64validator.AtLeast(5)` on `idle_timeout_minutes`, `AtLeast(1)` on memory; cross-field `min_replica_memory_gb <= max_replica_memory_gb` and `idle_timeout_minutes requires idle_scaling` checks |
| 8 | Non-deterministic weekday JSON | `sort.Ints(intWeekdays)` before assigning to the API entry |
| 9 | Hour-range asymmetry | Description now has a dedicated "Hour ranges" section with the 24-hour / overnight-wrap conventions; attribute descriptions call out the 0-23 vs 1-24 asymmetry |
| 10 | Hard-coded `10` | Now sourced from `api.MaxAutoScalingScheduleEntries` (exported from `pkg/internal/api/scheduled_scaling.go`) |
| 11 | Weekday test assertion checks length only | Now asserts sorted contents (`[1 2 3 4 5]`); resource test also asserts the sort-before-send behaviour with a deliberately unsorted input |
| 12 | Test coverage gaps | Added `TestApplyScheduleToState_PreservesMultiEntryOrder`, `TestRoundTrip_NoServerNormalization`, expanded `TestValidateScheduledScalingEntries` with memory-min>max + idle-timeout-without-idle-scaling cases. `ImportState` is exercised indirectly through the new top-level `id` write path; a Plugin Framework `resource.TestStep`-style import test still needs an acceptance harness (out of scope for this unit-test pass) |
| 13 | Empty-entries API test only proves serialization | Acknowledged; kept as a client-only safety net since the resource layer now blocks empty input via `SizeAtLeast(1)` |

### Not yet done
- Manual `terraform apply` + no-op-plan against a dev org with `canUseScheduledAutoscaling` enabled. This is the unchecked box in the PR description; the new round-trip test gives mechanical evidence but isn't a substitute. Will run before merge.

pkg/resource/service_scheduled_scaling.go
serdec: Fixed in 410adc7. All six optional entry fields are now `Optional + Computed + UseStateForUnknown`, matching `service.go:328-362`. Server-filled defaults will now flow into state and the next plan sees no diff.

pkg/resource/service_scheduled_scaling.go
serdec: Fixed in 410adc7 — added `listvalidator.SizeAtLeast(1)` on `entries`. The empty-list path in `planEntriesToAPI` is now unreachable from a real plan but kept as a client-only safety guard.

pkg/resource/service_scheduled_scaling.go
serdec: Fixed in 410adc7. Added `id` (Computed, `UseStateForUnknown`, = `service_id`). `ImportState` now writes both `id` and `service_id` (custom handler since they're equal but both need populating).

pkg/resource/service_scheduled_scaling.go
serdec: Fixed in 410adc7 — dropped the entry-level `id` field entirely. It wasn't user-facing and the "known after apply" noise wasn't worth keeping it. The PR description's claim about positional ID stability is irrelevant once the field is gone.

pkg/resource/service_scheduled_scaling.go
serdec: Fixed in 410adc7. `Create` now `GET`s first; if a non-empty schedule exists it errors with a 'please import' message including the exact `terraform import` command.

pkg/resource/service_scheduled_scaling.go
serdec: Fixed in 410adc7. `Create` and `Update` now detect 404 and surface 'Service not found' with a hint to check the `service_id` reference, instead of the raw `status: 404, body: ...`.

pkg/resource/service_scheduled_scaling.go
serdec: Fixed in 410adc7. Added `int64validator.AtLeast(5)` on `idle_timeout_minutes`, `AtLeast(1)` on memory, and cross-field checks for `min_replica_memory_gb <= max_replica_memory_gb` and `idle_timeout_minutes requires idle_scaling = true`. New test cases cover both cross-field rules.

pkg/resource/service_scheduled_scaling.go
serdec: Fixed in 410adc7 — `sort.Ints(intWeekdays)` before assigning to the API entry. Added a test using a deliberately unsorted `[3, 1]` set as input and asserting `[1, 3]` on the wire.

pkg/resource/service_scheduled_scaling.go
serdec: Fixed in 410adc7. Description now has a dedicated 'Hour ranges' section spelling out the 0-23 vs 1-24 asymmetry and the 24-hour-window / overnight-wrap conventions. Attribute descriptions also call this out at the field level.

pkg/resource/service_scheduled_scaling.go
serdec: Fixed in 410adc7. The constant now lives in `pkg/internal/api/scheduled_scaling.go` as `MaxAutoScalingScheduleEntries = 10` and the resource references it. One place to bump if the server raises the cap.

pkg/resource/service_scheduled_scaling_test.go
serdec: Fixed in 410adc7. The test now extracts weekdays, sorts, and asserts `[1, 2, 3, 4, 5]` via `reflect.DeepEqual`.

pkg/resource/service_scheduled_scaling_test.go
serdec: Partially addressed in 410adc7. Added: `TestApplyScheduleToState_PreservesMultiEntryOrder` (multi-entry order), `TestRoundTrip_NoServerNormalization` (plan → API → state matches plan), and more `TestValidateScheduledScalingEntries` cases. ImportState is exercised mechanically via the new top-level `id` write path, but a full Plugin Framework import `TestStep` requires an acceptance-test harness — flagged for follow-up. The manual `terraform apply` checkbox in the PR description is still outstanding and remains the most valuable verification.

pkg/internal/api/scheduled_scaling_test.go
serdec: Acknowledged in 410adc7. The resource layer now rejects empty input via `SizeAtLeast(1)`, so the empty-entries POST is dead code in practice. Kept the API client test as a defense-in-depth check on the wire layer.

serdec (commented):
Second pass on commit 410adc7. Round 1 feedback is well-addressed — top-level `id` added, `SizeAtLeast(1)`, Optional+Computed+UseStateForUnknown on all defaultable fields, pre-create check, sorted weekdays, expanded validation, multi-entry order test, etc.

New comments below are minor. The one thing still load-bearing: **the manual apply (test plan checkbox) is still unchecked**, and the unit-level round-trip test added doesn't actually exercise the drift case I was worried about — see inline comment on `TestRoundTrip_NoServerNormalization`. That test verifies the model mapping is faithful when both sides have identical data, but it doesn't simulate the server filling in fields the user didn't set — which is the only scenario where the schema fix (Optional+Computed) matters. A `terraform apply` followed by a no-op `terraform plan` against a real server with a partially-specified entry remains the strongest verification.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,577 @@
...
+	if resp.Diagnostics.HasError() {
+		return
+	}
+
+	serviceID := state.ServiceID.ValueString()
+	if serviceID == "" {
+		// Populated from `id` on import.
+		serviceID = state.ID.ValueString()
+		state.ServiceID = state.ID
+	}
```
serdec: Dead code now. `ImportState` (line 337-338) writes both `id` and `service_id`, and `Create`/`Update` both set `plan.ID = plan.ServiceID` before persisting. There's no execution path where `state.ServiceID` is empty in Read.

Safe to delete this fallback. If you want a belt-and-braces assertion, replace with:

```go
if state.ServiceID.IsNull() || state.ServiceID.ValueString() == "" {
    resp.Diagnostics.AddError("Invalid state", "service_id is empty — this resource may have been corrupted by a partial import.")
    return
}
```

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,577 @@
...
+		resp.Diagnostics.AddError("Error checking for existing scheduled scaling", err.Error())
+		return
+	}
+	if existing != nil && len(existing.Entries) > 0 {
+		resp.Diagnostics.AddError(
+			"Scheduled scaling already exists for this service",
+			fmt.Sprintf("Service %s already has a scaling schedule with %d entries. Import it into Terraform with: terraform import clickhouse_service_scheduled_scaling.<name> %s", serviceID, len(existing.Entries), serviceID),
+		)
+		return
+	}
```
serdec, line 231: Subtle design decision worth a code comment: the check `existing != nil && len(existing.Entries) > 0` treats a 200-response with empty entries as "no schedule exists", not as "a schedule with zero entries exists". This is the right call (an empty schedule is observationally equivalent to no schedule), and it makes the provider robust to whichever shape the server returns for "no schedule" — 404 or 200+empty. Worth one line saying so, since it's the kind of thing that looks like a bug at a glance.

Also: no unit test covers this code path. Easy to add given there's already a ClientMock — three test cases (`GET returns 404 → Create proceeds`, `GET returns 200+empty → Create proceeds`, `GET returns 200+entries → Create errors with import hint`).

pkg/resource/service_scheduled_scaling_test.go
```diff
@@ -0,0 +1,439 @@
...
+	}
+	got := make([]int, len(wd))
+	for i, v := range wd {
+		got[i] = int(v)
+	}
+	sort.Ints(got)
+	if !reflect.DeepEqual(got, []int{1, 2, 3}) {
+		t.Errorf("weekdays = %v; want [1 2 3]", got)
+	}
+}
```
serdec, line 504: This test doesn't actually exercise the drift case. The plan has every optional field set, so `planEntriesToAPI` produces a fully-populated request, the test then echoes that back verbatim, and `applyScheduleToState` of course reproduces the same model. It proves the *mapping* is faithful, but it can't catch the original concern: server fills in fields the user *didn't* set.

More useful complement (call it `TestRoundTrip_ServerFillsDefaults`):

```go
// User omits idle_scaling / idle_timeout / memory bounds; server fills them.
planEntry := models.ScheduledScalingEntryModel{
    Name:         types.StringValue("sparse"),
    Weekdays:     weekdaySet,
    StartHourUtc: types.Int64Value(9),
    EndHourUtc:   types.Int64Value(17),
    MinReplicas:  types.Int64Value(2),
    MaxReplicas:  types.Int64Value(2),
    // idle_scaling, idle_timeout_minutes, memory bounds: not set
}
apiEntries, _ := planEntriesToAPI(ctx, buildEntryList(t, planEntry))

// Simulate the server filling in defaults the caller didn't send.
serverResponse := &api.AutoScalingSchedule{Entries: []api.AutoScalingScheduleEntry{{
    ...apiEntries[0]...
    MinReplicaMemoryGb: intPtr(8),    // server default
    MaxReplicaMemoryGb: intPtr(8),    // server default
    IdleScaling:        boolPtr(true), // server default
    IdleTimeoutMinutes: intPtr(10),   // server default
}}}

// Map back to state, then assert state has the server-filled values.
// This proves the model can carry server defaults without panicking.
```

The schema-level guarantee (no perpetual drift via `UseStateForUnknown`) still needs an acceptance test or manual `terraform plan` against a live server — neither this test nor the framework can prove it at the unit level.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,577 @@
...
+		if !e.IdleTimeoutMinutes.IsNull() && !e.IdleTimeoutMinutes.IsUnknown() {
+			idleSet := !e.IdleScaling.IsNull() && !e.IdleScaling.IsUnknown()
+			if !idleSet || !e.IdleScaling.ValueBool() {
+				diags.AddAttributeError(
+					entryPath,
+					"idle_timeout_minutes requires idle_scaling = true",
+					"idle_timeout_minutes only has an effect when idle_scaling is enabled for this entry.",
+				)
+			}
+		}
```
serdec: Stricter than `service.go:359-363`'s `int64validator.AlsoRequires(MatchRoot("idle_scaling"))`. That validator only requires `idle_scaling` to be *present* — your check requires `idle_scaling = true` (value, not presence).

Arguably better, and the friendlier error is consistent with this PR's design. But it does mean a config that's valid against `clickhouse_service` (`idle_scaling=false, idle_timeout=10`, where the timeout is silently ignored) will *fail* validation here. Worth a single-line acknowledgment in the entry's `idle_timeout_minutes` schema description: "Setting this with `idle_scaling = false` is rejected" — so users porting from `clickhouse_service` aren't surprised.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,577 @@
...
+	if err != nil && !api.IsNotFound(err) {
+		resp.Diagnostics.AddError("Error deleting scheduled scaling", err.Error())
+	}
+}
+
+func (r *ServiceScheduledScalingResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	// `id` and `service_id` are equal — write both so Read finds the service.
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), req.ID)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("service_id"), req.ID)...)
+}
```
serdec, line 341: Edge: an empty `req.ID` writes empty strings to both attributes. Read then calls `GetScheduledScaling(ctx, "")` which the server will probably reject with a confusing error (the API URL becomes `/services//scalingSchedule`).

Cheap guard:

```go
if req.ID == "" {
    resp.Diagnostics.AddError("Invalid import ID", "Provide the service ID: terraform import clickhouse_service_scheduled_scaling.<name> <service_id>")
    return
}
```

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,577 @@
...
+				"Service not found",
+				fmt.Sprintf("Service %s no longer exists. Remove the resource from configuration or recreate the service.", plan.ServiceID.ValueString()),
+			)
+			return
+		}
+		resp.Diagnostics.AddError("Error updating scheduled scaling", err.Error())
+		return
+	}
+
+	plan.ID = plan.ServiceID
```
serdec, line 315: Nit: `plan.ID = plan.ServiceID` in Update is redundant. `service_id` has `RequiresReplace()`, so Update is never called when `service_id` changes. Combined with `UseStateForUnknown` on `id`, the framework has already preserved the prior `id` value (= prior `service_id` = current `service_id`). Cosmetic, harmless. If you want to keep an explicit write for clarity, fine.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,577 @@
...
+func (r *ServiceScheduledScalingResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		MarkdownDescription: serviceScheduledScalingResourceDescription,
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Description: "Resource identifier. Equal to service_id (one schedule per service).",
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
```
serdec, line 59: Tiny doc nit: "Equal to service_id (one schedule per service)" — would help to add "Changes to `service_id` replace the resource." Users grepping for `id` semantics will appreciate the explicit pointer to `service_id`'s `RequiresReplace`.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,577 @@
...
+	err := r.client.DeleteScheduledScaling(ctx, state.ServiceID.ValueString())
+	if err != nil && !api.IsNotFound(err) {
+		resp.Diagnostics.AddError("Error deleting scheduled scaling", err.Error())
+	}
+}
+
+func (r *ServiceScheduledScalingResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	// `id` and `service_id` are equal — write both so Read finds the service.
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), req.ID)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("service_id"), req.ID)...)
```
serdec, line 340: Edge: an empty `req.ID` writes empty strings to both attributes. Read then calls `GetScheduledScaling(ctx, "")` and the server URL becomes `/services//scalingSchedule` — a confusing error.

Cheap guard:

```go
if req.ID == "" {
    resp.Diagnostics.AddError("Invalid import ID", "Provide the service ID: terraform import clickhouse_service_scheduled_scaling.<name> <service_id>")
    return
}
```

pkg/resource/service_scheduled_scaling.go
serdec, line 315: Nit: `plan.ID = plan.ServiceID` here is redundant. `service_id` has `RequiresReplace()` so Update is never called when it changes; combined with `UseStateForUnknown` on `id`, the framework has already preserved the prior `id` value (= prior `service_id` = current `service_id`). Cosmetic — harmless if kept for clarity.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,577 @@
...
+func (r *ServiceScheduledScalingResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_service_scheduled_scaling"
+}
+
+func (r *ServiceScheduledScalingResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
+	resp.Schema = schema.Schema{
+		MarkdownDescription: serviceScheduledScalingResourceDescription,
+		Attributes: map[string]schema.Attribute{
+			"id": schema.StringAttribute{
+				Description: "Resource identifier. Equal to service_id (one schedule per service).",
```
serdec, line 55: Tiny doc nit: "Equal to service_id (one schedule per service)" — worth adding "Changes to `service_id` replace the resource." Users grepping for `id` semantics appreciate the explicit pointer to `service_id`'s `RequiresReplace`.

serdec:
Cleaned up the two items I'd left outstanding in the previous reply.

### ImportState test (comment [#3232720626](https://github.com/ClickHouse/terraform-provider-clickhouse/pull/536#discussion_r3232720626))

Added `TestServiceScheduledScalingResource_ImportState` in [12d2e41](https://github.com/ClickHouse/terraform-provider-clickhouse/pull/536/commits/12d2e41). It drives the actual `ImportState` handler through a `tfsdk.State` constructed from the resource's own `Schema` and asserts both `id` and `service_id` get populated from the import ID. No acceptance harness needed.

```go
r.ImportState(ctx, resource.ImportStateRequest{ID: "abc-123"}, resp)
// resp.State.GetAttribute(..., "id")          -> "abc-123"
// resp.State.GetAttribute(..., "service_id") -> "abc-123"
```

### Smoke validation against the example HCL

Built the provider locally, wired it in via `dev_overrides`, and ran `terraform validate` on three configurations:

**1. The full example from `examples/resources/clickhouse_service_scheduled_scaling/resource.tf` + a wrap-overnight entry + a 24-hour window entry:**

```
Success! The configuration is valid
```

**2. Negative case — `idle_timeout_minutes` without `idle_scaling=true`, and `start_hour_utc == end_hour_utc`:**

```
Error: start_hour_utc and end_hour_utc must differ
       Zero-duration windows are not allowed.
Error: idle_timeout_minutes requires idle_scaling = true
       idle_timeout_minutes only has an effect when idle_scaling is enabled for this entry.
```

**3. Negative case — empty `entries = []`:**

```
Error: Invalid Attribute Value
       Attribute entries list must contain at least 1 elements, got: 0
```

The full live-apply round-trip against a dev org with `canUseScheduledAutoscaling` still requires real credentials and a real service, which is out of my scope. I've left the checkbox in the PR description unchecked so a maintainer can run it before merge. The new tests + `terraform validate` give mechanical evidence that:
- the schema accepts every realistic configuration,
- every cross-field validator fires with the user-facing wording I authored,
- `SizeAtLeast(1)` blocks empty-entries configurations at plan time,
- the ImportState handler populates both keys.

serdec:
Fixed in [c939b95](https://github.com/ClickHouse/terraform-provider-clickhouse/pull/536/commits/c939b95).

The validator I added in 410adc7 was stricter than the server. The server (and the UI) accepts entries with `idle_scaling = false` (or unset) and `idle_timeout_minutes` set, which broke the UI-create-then-import-into-HCL flow: state was correct but any HCL mirroring it was rejected at plan/validate time.

Now using `AddAttributeWarning` instead of `AddAttributeError`. The user still gets a heads-up that the field has no effect for that entry, but the import path works.

Verified locally with the exact repro you described:

```hcl
entries = [{
  name                 = "ui-persisted"
  weekdays             = [1, 2, 3]
  start_hour_utc       = 8
  end_hour_utc         = 18
  idle_scaling         = false
  idle_timeout_minutes = 15
}]
```

`terraform validate` now emits:

```
Warning: idle_timeout_minutes set without idle_scaling = true
  …
  idle_timeout_minutes only has an effect when idle_scaling is enabled.
  The server accepts this combination; this is just a heads-up that the
  timeout will not take effect for this entry.

Success! The configuration is valid, but there were some validation warnings
```

Test coverage updated: `TestValidateScheduledScalingEntries/idle_timeout_*` cases now assert `ErrorsCount=0, WarningsCount=1`.

Thanks for catching this — the original reviewer's recommendation ("client-side rejection is friendlier") didn't anticipate that the server is more permissive than the suggested validator.

serdec:
You're right — the warning message was wrong. Fixed in [2715d95](https://github.com/ClickHouse/terraform-provider-clickhouse/pull/536/commits/2715d95).

The validator went through two wrong shapes before this:

1. **410adc7**: errored when `idle_scaling != true`. Wrong because the server accepts `idle_scaling=false` as long as `idle_timeout_minutes` is also set (the UI repro case).
2. **c939b95**: downgraded to a warning with the message "server accepts this combination". That message claimed the server is permissive when in fact it requires the pair.

Now it's a **pair-required error**, the same pattern as the existing `min_replica_memory_gb` / `max_replica_memory_gb` rule and `min_replicas` / `max_replicas` rule. Both fields together, or neither — matches what the server actually enforces.

Verified locally with `terraform validate`:

- **UI-persisted repro** (`idle_scaling = false` + `idle_timeout_minutes = 15`) → `Success! The configuration is valid` ✅
- **Lone `idle_timeout_minutes`** → `Error: idle_scaling and idle_timeout_minutes must be set together` ✅
- **Lone `idle_scaling`** → same error ✅
- **Both unset** → `Success!` ✅

Test cases updated: `TestValidateScheduledScalingEntries` now covers all four quadrants (`idle_scaling_without_idle_timeout`, `idle_timeout_without_idle_scaling`, `idle_scaling=false_with_idle_timeout_set_is_accepted`, `both_idle_fields_unset`).

serdec:
Good push — I'd been guessing at server behavior. Read the actual server code this time:

- **`validateScheduleEntryPublicFields`** ([`apps/openapi/src/v1/AutoScalingScheduleV1Handler.ts:83-112`](https://github.com/ClickHouse/control-plane/blob/main/apps/openapi/src/v1/AutoScalingScheduleV1Handler.ts#L83-L112))
- **`validateScheduleEntries`** ([`packages/cp-common-server/src/instance/AutoScalingScheduleManager.ts:118-237`](https://github.com/ClickHouse/control-plane/blob/main/packages/cp-common-server/src/instance/AutoScalingScheduleManager.ts#L118-L237))

Neither layer has any cross-field rule between `idleScaling` and `idleTimeoutMinutes`. Each is independently `undefinedOr(...)`. Server accepts **all four** combinations: both set (with either boolean for `idleScaling`), only one set, neither set.

Fixed in [7956a29](https://github.com/ClickHouse/terraform-provider-clickhouse/pull/536/commits/7956a29) — dropped the client-side check entirely. Three wrong shapes preceded this:

| Commit | Shape | Wrong because |
|---|---|---|
| 410adc7 | Errored when `idle_scaling != true` | Server accepts `idle_scaling=false` |
| c939b95 | Warning saying "server accepts this combination" | Wrong message and overly chatty |
| 2715d95 | Pair-required error | Server doesn't require pairing |
| 7956a29 | **No check** | ✅ matches server |

Verified all five permutations with `terraform validate` against the locally built provider:

```hcl
{ idle_scaling = true,  idle_timeout_minutes = 10 } # both set
{ idle_scaling = false, idle_timeout_minutes = 15 } # UI repro
{                       idle_timeout_minutes = 5  } # lone timeout
{ idle_scaling = true                             } # lone scaling
{                                                 } # neither
```

→ `Success! The configuration is valid` for all five.

Test coverage: `TestValidateScheduledScalingEntries` now locks in all four idle combinations as accepted (no errors, no warnings) so this regression can't sneak back in.

The other validators in this helper (paired memory + min ≤ max, paired replicas + min == max, start ≠ end) all map 1:1 to assertions in the server's `validateScheduleEntries` / `validateScheduleEntryPublicFields`, so those stay as errors. Sorry for the multiple iterations — should have read the server code first instead of trusting iteratively-stated assumptions.

examples/resources/clickhouse_service_scheduled_scaling/resource.tf
```diff
@@ -0,0 +1,25 @@
...
+      weekdays             = [0, 1, 2, 3, 4, 5, 6]
+      start_hour_utc       = 22
+      end_hour_utc         = 6
+      min_replicas         = 1
+      max_replicas         = 1
+      idle_scaling         = true
+      idle_timeout_minutes = 5
+    },
+  ]
+}
```
nunoadrego, line 29: Similar to existing service-related resources:
- https://github.com/ClickHouse/terraform-provider-clickhouse/blob/main/examples/resources/clickhouse_service_private_endpoints_attachment/resource.tf
- https://github.com/ClickHouse/terraform-provider-clickhouse/blob/main/examples/resources/clickhouse_service_transparent_data_encryption_key_association/resource.tf

```suggestion
resource "clickhouse_service" "svc" {
  ...
}

resource "clickhouse_service_scheduled_scaling" "example" {
  service_id = clickhouse_service.svc.id

  entries = [
    {
      name           = "Business hours"
      weekdays       = [1, 2, 3, 4, 5]
      start_hour_utc = 8
      end_hour_utc   = 18
      min_replicas   = 3
      max_replicas   = 3
      idle_scaling   = false
    },
    {
      name                 = "Overnight"
      weekdays             = [0, 1, 2, 3, 4, 5, 6]
      start_hour_utc       = 22
      end_hour_utc         = 6
      min_replicas         = 1
      max_replicas         = 1
      idle_scaling         = true
      idle_timeout_minutes = 5
    },
  ]
}
```

pkg/resource/descriptions/service_scheduled_scaling.md
```diff
@@ -0,0 +1,51 @@
+You can use the *clickhouse_service_scheduled_scaling* resource to manage time-based scaling rules for a ClickHouse Cloud service.
+
+A schedule is a list of recurring weekly windows. While a window is active the service uses the replica count, memory bounds, and idle-scaling settings declared on that entry; otherwise the service falls back to its base auto-scaling configuration.
+
+~> **Note:** This resource is in beta. Scheduled scaling must be enabled for your organization (`canUseScheduledAutoscaling`). Reach out to ClickHouse support if the API returns `403 FORBIDDEN`. The server currently requires `min_replicas == max_replicas` per entry, and a maximum of 10 entries per schedule.
```
nunoadrego, line 5: Shall we add this resource as beta in the provider? If so, remove resource from - `pkg/resource/register_stable.go`

pkg/resource/service_scheduled_scaling.go
```diff
@@ -0,0 +1,571 @@
...
+						},
+						"end_hour_utc": schema.Int64Attribute{
+							Description: "End hour in UTC (1-24). Must differ from start_hour_utc. Note the asymmetric range: end_hour_utc=0 is invalid; use end_hour_utc=24 to mean midnight at end of day.",
+							Required:    true,
+							Validators: []validator.Int64{
+								int64validator.Between(1, 24),
+							},
+						},
+						"min_replica_memory_gb": schema.Int64Attribute{
+							Description: "Minimum memory per replica in GiB. Must be set together with max_replica_memory_gb.",
```
nunoadrego, line 108: We can use something similar to:
`stringvalidator.AlsoRequires(path.Expressions{path.MatchRoot("max_replica_memory_gb")}...),`

And vice versa in `max_replica_memory_gb`.

pkg/resource/service_scheduled_scaling.go
nunoadrego: This comment seems relevant

pkg/resource/service_scheduled_scaling.go
nunoadrego: I saw you added the guardrails in validateScheduledScalingEntries, but if you add them here, it fails earlier for the user.

nunoadrego (commented):
Left some comments. Can you also check the following:
- run make docs
- add entry in `examples/full` for e2e tests

serdec:
Thanks for the review! Addressed in [36c3a43](https://github.com/ClickHouse/terraform-provider-clickhouse/pull/536/commits/36c3a43). Per-comment replies inline; recap below.

### Inline comments

| Comment | Resolution |
|---|---|
| Example should use a `clickhouse_service` block for `service_id` | Done. Mirrors the `service_private_endpoints_attachment` / `tde_key_association` examples. |
| Mark resource as beta → remove from `register_stable.go` | Done. Resource is now alpha-only via `register_debug.go`. `make docs` no longer emits the stable doc; `make docs-alpha` regenerates it (same pattern as `role` / `role_assignment`). |
| Use `int64validator.AlsoRequires` for the memory/replica pair-required rules | Done. Added on `min_replica_memory_gb` ↔ `max_replica_memory_gb` and `min_replicas` ↔ `max_replicas` using `path.MatchRelative().AtParent().AtName(...)`. Dropped the corresponding "set together" branches from `validateScheduledScalingEntries`. The `min <= max` (memory) and `min == max` (replicas) checks stay in `ValidateConfig` since they're true cross-field comparisons. |
| "This comment seems relevant" — dead `if serviceID == ""` branch in `Read()` | Done. `ImportState` already writes both `id` and `service_id`, so the fallback was unreachable. Removed. |
| Earlier-failure point for guardrails | Folded into the `AlsoRequires` move above. |

### Top-level checklist

- **`make docs`** — ran. The stable doc for `service_scheduled_scaling` is gone (resource is alpha-only); `make docs-alpha` regenerates it, matching the existing alpha-doc lifecycle managed by the release bot.
- **`examples/full` entry** — added `examples/full/scheduled_scaling/aws/` (`main.tf`, `provider.tf`, `provider.tf.template`, `variables.tfvars.sample`, `README.md`). `list-examples` auto-enrols it in the e2e matrix. Note: needs `canUseScheduledAutoscaling` enabled on whichever org runs the e2e job.

### Verification

- `go build ./...` and `go build -tags alpha ./...` clean.
- `go vet` clean both modes.
- `go test ./...` and `go test -tags alpha ./...` pass; `TestValidateScheduledScalingEntries` cases for the pair-mismatch scenarios were removed since the schema-level validator catches them earlier (and a unit-test harness around the schema-level `AlsoRequires` would mostly be testing the framework).

examples/resources/clickhouse_service_scheduled_scaling/resource.tf
serdec: Done in 36c3a43. The example now declares a `clickhouse_service.svc` block and wires `service_id = clickhouse_service.svc.id`, matching the convention from the two examples you linked.

pkg/resource/descriptions/service_scheduled_scaling.md
serdec: Done in 36c3a43 — went with beta. Removed `NewServiceScheduledScalingResource` from `pkg/resource/register_stable.go`; it stays in `register_debug.go` so it's only registered for `-tags alpha` builds. `make docs` no longer emits `docs/resources/service_scheduled_scaling.md`; `make docs-alpha` does, matching the lifecycle the release bot already manages for `role` / `role_assignment`.

pkg/resource/service_scheduled_scaling.go
serdec: Done in 36c3a43. Both pairs now use `int64validator.AlsoRequires(path.MatchRelative().AtParent().AtName("..."))`:

- `min_replica_memory_gb` ↔ `max_replica_memory_gb`
- `min_replicas` ↔ `max_replicas`

`path.MatchRelative().AtParent().AtName(...)` because the attributes live inside the nested entry object (same pattern as `clickpipe.go:355`). The "set together" branches were removed from `validateScheduledScalingEntries` — the schema-level validator now catches that case before `ValidateConfig` runs. The `min <= max` (memory) and `min == max` (replicas) checks stay there since they're true cross-field value comparisons.

pkg/resource/service_scheduled_scaling.go
serdec: Agreed — removed in 36c3a43. The `ImportState` handler writes both `id` and `service_id`, so by the time `Read` runs `state.ServiceID` is always populated. The `if serviceID == ""` branch was unreachable. `Read` now just reads `state.ServiceID.ValueString()` directly.

pkg/resource/service_scheduled_scaling.go
serdec: Done in 36c3a43 — folded into the `AlsoRequires` change above. The "set together" guardrails for the two pairs are now schema-level via `int64validator.AlsoRequires`, so the error surfaces at schema-validation time (before `ValidateConfig`) and includes the offending attribute path automatically. The corresponding branches in `validateScheduledScalingEntries` were removed; only the cross-field value comparisons (`min <= max` for memory, `min == max` for replicas) and the `start != end` window check remain there.

serdec:
Done in [883ecf0](https://github.com/ClickHouse/terraform-provider-clickhouse/pull/536/commits/883ecf0). Verified the non-overlap invariant in `packages/cp-common-server/src/instance/AutoScalingScheduleManager.ts:228-236` — the server rejects any pair of entries that overlap on day+hour, so at most one window applies at a time and entry order is purely cosmetic.

Switched to `SetNestedAttribute`:

- Schema: `ListNestedAttribute` → `SetNestedAttribute`; `listvalidator.SizeAtLeast/AtMost` → `setvalidator.*`.
- Model: `types.List` → `types.Set` on `Entries`.
- Conversion helpers (`planEntriesToAPI`, `applyScheduleToState`) updated; `ElementsAs` works identically on both.
- `validateScheduledScalingEntries` no longer indexes via `AtListIndex` — the offending entry name now appears in the diagnostic detail instead.
- Description text dropped the "ordered list / first match wins" framing.
- Multi-entry-order test was repurposed into a set-membership check.

HCL syntax is unchanged (`entries = [ {...}, {...} ]` — Terraform coerces the tuple into a set). Reordering the entries in HCL now produces "No changes" instead of a per-entry diff.

nunoadrego (approved):

PR #544, opened 2026-05-20
feat(api): add Managed Postgres HTTP client

amogiska:
## Summary

- Hand-written HTTP client for the ClickHouse Cloud Managed Postgres endpoints under `/v1/organizations/{orgId}/postgres`. Mirrors the existing `service.go` / `clickpipe.go` pattern: models in `postgres_models.go`, methods on `*ClientImpl` in `postgres.go`, interface extension in `interface.go`, regenerated `client_mock.go`.
- **13 client methods**: `GetPostgres`, `ListPostgres`, `CreatePostgres`, `UpdatePostgres`, `DeletePostgres`, `RestorePostgres`, `CreatePostgresReadReplica`, `SetPostgresPassword`, `GetPostgresConfig`, `ReplacePostgresConfig`, `GetPostgresCaCertificates`, plus two wait helpers: `WaitForPostgresState` (mirrors `WaitForServiceState`) and `WaitForPostgresStateTransitionAndReturn` (round-trip wait for post-PATCH transitions).
- `DeletePostgres` retries 409 conflicts for ~15 min but fails fast when the response indicates a dependent replica blocks deletion.
- `PgConfigMap` has a custom `UnmarshalJSON` (via `json.Decoder.UseNumber()` + type switch) that accepts both string and numeric values, matching the server's `[key: string]: string | number` wire shape. Rejects bool / null / object / array values.
- `PostgresUpdate.Tags` is `*[]Tag` so callers can distinguish "leave tags alone" (nil) from "clear all tags" (`&[]Tag{}`). Other fields default to plain types, matching the upstream OpenAPI contract.
- **Sensitive-field redaction** in `common.go:doRequest`: `password`, `newPassword`, `newPasswordHash`, `newDoubleSha1Hash`, `password_wo`, `tokenSecret`, `connectionString` (URI embeds password), and `secrets` / `credentials` containers are stamped to `"REDACTED"` before going to `tflog`. Fixes a pre-existing leak where `TF_LOG=DEBUG` during a `clickhouse_service` password rotation logged the plaintext hash.
- Adds `@ClickHouse/clickgres` as CODEOWNER for `**/postgres*` files (mirrors the existing `**/clickpipe*` pattern).

## Test plan

- [x] `go test ./pkg/internal/api/...` — 48 new top-level tests (29 client + 17 model + 2 redaction), 60 PASS lines including subtests
- [x] `go test ./...` — whole module green (116 PASS, 0 FAIL across 4 packages)
- [x] `go vet ./...` clean
- [x] `golangci-lint run ./...` (pinned `v1.64.8` from Makefile) clean
- [x] `make mock` regenerated `client_mock.go` cleanly; all 13 Postgres methods present
- [x] Pre-commit hook (`make fmt && make docs && make build`) green on every commit


pkg/internal/api/client_mock.go
amogiska, line 1: All of this is generated

pkg/internal/api/common.go
```diff
@@ -20,6 +20,97 @@ type ResponseWithResult[T any] struct {
...
+			if _, ok := sensitiveBodyContainers[k]; ok {
+				t[k] = redactedPlaceholder
+				continue
+			}
+			t[k] = redactJSONValue(child)
+		}
+		return t
+	case []interface{}:
+		for i, child := range t {
+			t[i] = redactJSONValue(child)
```
theory, line 105: I presume this is to be conservative in that we don't have any arrays rn, yes?

pkg/internal/api/common.go
```diff
@@ -112,17 +200,7 @@ func (c *ClientImpl) doRequest(ctx context.Context, initialReq *http.Request) ([
...
-			err = json.Indent(&buf, body, "", "  ")
-			if err != nil {
-				// Parsing/indentation failed, fallback to raw body
-				debugctx = tflog.SetField(debugctx, "responseBody", string(body))
-			} else {
-				// Parsing ok, use formatted body.
-				debugctx = tflog.SetField(debugctx, "responseBody", buf.String())
-			}
-		}
+		debugctx = tflog.SetField(debugctx, "responseBody", formatLogBody(body))
```
theory: Hell yeah de-duping.

pkg/internal/api/common.go
```diff
@@ -20,6 +20,97 @@ type ResponseWithResult[T any] struct {
...
+				continue
+			}
+			if _, ok := sensitiveBodyContainers[k]; ok {
+				t[k] = redactedPlaceholder
+				continue
+			}
+			t[k] = redactJSONValue(child)
+		}
+		return t
+	case []interface{}:
```
theory: Prefer the newer and nicer-to-read `any` (unless the rest of the code base is littered with `[]inteface{}`s.

pkg/internal/api/common_test.go
```diff
@@ -0,0 +1,248 @@
...
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestRedactSensitiveBody(t *testing.T) {
+	cases := []struct {
```
theory, line 13: No tests for slices that I can see. Also worthwhile throwing in bools, numbers and nulls to make sure they're all passed through properly.

pkg/internal/api/common_test.go
```diff
@@ -0,0 +1,248 @@
...
+		},
+		{
+			name:  "non-object JSON returned as-is when no sensitive content",
+			input: `["a","b","c"]`,
+			want:  `["a","b","c"]`,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
```
theory, line 131: Consider using `t.Parallel()` in all tests so accelerate testing.

pkg/internal/api/common_test.go
```diff
@@ -0,0 +1,248 @@
...
+		secret := "Hunter2-Aa1!"
+		input := `{"result":{"id":"pg-1","password":"` + secret + `","connectionString":"postgresql://default:` + secret + `@host:5432/db?channel_binding=require"}}` //nolint:gosec // synthetic test fixture
+		got := formatLogBody([]byte(input))
+		if jsonContainsValueAnywhere(t, got, secret) {
+			t.Errorf("plaintext secret leaked through connectionString: %s", got)
+		}
+	})
+}
+
+// jsonContainsPath returns true iff the JSON document contains the (key, value)
```
theory:
```suggestion
// jsonContainsPath returns true if the JSON document contains the (key, value)
```

pkg/internal/api/common_test.go
```diff
@@ -0,0 +1,248 @@
...
+	if err := json.Unmarshal([]byte(doc), &v); err != nil {
+		t.Fatalf("doc not valid JSON: %v", err)
+	}
+	return walkFind(v, func(k string, val interface{}) bool {
+		s, ok := val.(string)
+		return ok && k == key && s == value
+	})
+}
+
+// jsonContainsValueAnywhere returns true iff the substring appears in the
```
theory:
```suggestion
// jsonContainsValueAnywhere returns true if the substring appears in the
```

pkg/internal/api/common_test.go
```diff
@@ -0,0 +1,248 @@
...
+// serialized form of the document. Useful for sanity checks ("did the secret
+// leak through any path").
+func jsonContainsValueAnywhere(t *testing.T, doc string, needle string) bool {
+	t.Helper()
+	return jsonStringContains(doc, needle)
```
theory: Odd to have this in a separate function since it's only ever used this once. But wouldn't it simpler to use [string.Contains](https://pkg.go.dev/strings#Contains)?

pkg/internal/api/interface.go
```diff
@@ -59,4 +59,18 @@ type Client interface {
 	CreateRole(ctx context.Context, req RoleCreateRequest) (*RBACRole, error)
 	UpdateRole(ctx context.Context, roleId string, req RoleUpdateRequest) (*RBACRole, error)
 	DeleteRole(ctx context.Context, roleId string) error
+
+	GetPostgres(ctx context.Context, postgresId string) (*Postgres, error)
+	ListPostgres(ctx context.Context) ([]PostgresListItem, error)
```
theory, line 67: Ugh, really want this to end in a plural, but `List Postgreses` is gross and `ListPostgresServices` breaks the pattern of the other functions.

pkg/internal/api/interface.go
```diff
@@ -59,4 +59,18 @@ type Client interface {
...
+	UpdatePostgres(ctx context.Context, postgresId string, body PostgresUpdate) (*Postgres, error)
+	DeletePostgres(ctx context.Context, postgresId string) error
+	WaitForPostgresState(ctx context.Context, postgresId string, stateChecker func(string) bool, maxWaitSeconds int) error
+	WaitForPostgresLeaveAndReturn(ctx context.Context, postgresId string, terminalState string, maxWaitSeconds int) error
+	RestorePostgres(ctx context.Context, sourceId string, body PostgresRestoreRequest) (*Postgres, error)
+	SetPostgresPassword(ctx context.Context, postgresId string, body PostgresPassword) (*PostgresPassword, error)
+	CreatePostgresReadReplica(ctx context.Context, sourceId string, body PostgresReadReplicaRequest) (*Postgres, error)
+	GetPostgresConfig(ctx context.Context, postgresId string) (*PostgresConfig, error)
+	ReplacePostgresConfig(ctx context.Context, postgresId string, body PostgresConfig) (*PostgresConfigUpdateResponse, error)
+	GetPostgresCaCertificates(ctx context.Context, postgresId string) ([]byte, error)
```
theory, line 78: Please add a documentation comment for each of these to explain their purpose.

pkg/internal/api/postgres_models.go
```diff
@@ -0,0 +1,150 @@
...
+	PostgresStateDeleting          = "deleting"
+)
+
+// PgConfigMap mirrors the server's `pgConfig` / `pgBouncerConfig` shape
+// `{[key: string]: string | number}`. Marshals as a plain string map;
+// unmarshal accepts mixed string and numeric values and coerces numbers
+// to their JSON string form via json.Number.
+type PgConfigMap map[string]string
+
+func (m *PgConfigMap) UnmarshalJSON(data []byte) error {
```
theory, line 28: Wondering why not use `UseNumber`. The API is fine to take non-string values. Assuming there's a reason, please explain it here.

pkg/internal/api/postgres_models.go
```diff
@@ -0,0 +1,150 @@
...
+		}
+		return fmt.Errorf("pgConfig key %q: unsupported value type %s", k, string(v))
+	}
+	return nil
+}
+
+// Postgres mirrors PostgresInstanceV1 (ManagedPostgresV1.ts:60-70).
+// storageSize is omitted intentionally — DEPRECATED server-side.
+type Postgres struct {
+	Id               string  `json:"id,omitempty"`
```
theory: No `UUID` type used by the ClickHouse instanced TF, I presume.

pkg/internal/api/postgres_models.go
```diff
@@ -0,0 +1,150 @@
...
+	Size            string      `json:"size"`
+	PostgresVersion string      `json:"postgresVersion,omitempty"`
+	HaType          string      `json:"haType,omitempty"`
+	Tags            []Tag       `json:"tags,omitempty"`
+	PgConfig        PgConfigMap `json:"pgConfig,omitempty"`
+	PgBouncerConfig PgConfigMap `json:"pgBouncerConfig,omitempty"`
+}
+
+// PostgresUpdate is the PATCH /postgres/{id} body. Server accepts ONLY
+// size / haType / tags; `name` is intentionally absent (no field in the
```
theory, line 101: This may change if we decide to merge ClickHouse/control-plane#33495.

pkg/internal/api/postgres_models_test.go
```diff
@@ -0,0 +1,256 @@
...
+func TestPgConfigMap_UnmarshalsStringValues(t *testing.T) {
+	input := []byte(`{"max_connections":"200","work_mem":"64MB"}`)
+	var got PgConfigMap
+	if err := json.Unmarshal(input, &got); err != nil {
+		t.Fatalf("unmarshal: %v", err)
+	}
+	want := PgConfigMap{"max_connections": "200", "work_mem": "64MB"}
+	if diff := cmp.Diff(want, got); diff != "" {
+		t.Errorf("PgConfigMap mismatch (-want +got):\n%s", diff)
+	}
```
theory, line 20: Good grief, does this project not use [testify](https://github.com/stretchr/testify)?

pkg/internal/api/postgres_models_test.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func TestPgConfigMap_UnmarshalsNumericValuesAsStrings(t *testing.T) {
```
theory, line 23: Underscores are generally frowned upon in Go function namers.

pkg/internal/api/postgres_models_test.go
```diff
@@ -0,0 +1,256 @@
...
+		t.Errorf("max_connections = %q; want 200", got["max_connections"])
+	}
+	if got["work_mem"] != "64" {
+		t.Errorf("work_mem = %q; want 64", got["work_mem"])
+	}
+}
+
+func TestPgConfigMap_UnmarshalsEmptyObject(t *testing.T) {
+	var got PgConfigMap
+	if err := json.Unmarshal([]byte(`{}`), &got); err != nil {
```
theory, line 58: Probably worth testing `null` and other scalars.

pkg/internal/api/postgres_models_test.go
```diff
@@ -0,0 +1,256 @@
...
+		t.Fatalf("marshal: %v", err)
+	}
+	if string(body) != `{}` {
+		t.Errorf("got %q; want {}", string(body))
+	}
+}
+
+func TestPostgresUpdate_DoesNotIncludeNameField(t *testing.T) {
+	// Server's PostgresInstancePatchRequestV1 has no `name` field; sending
+	// one would silently no-op. Guard against accidental addition.
```
theory: Is this not true for all fields in `PostgresUpdate`? I don't really understand what this test is for.

pkg/internal/api/postgres_models_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+func TestPostgresUpdate_NilTagsOmitsField(t *testing.T) {
+	body, err := json.Marshal(PostgresUpdate{Size: "r6gd.large"})
+	if err != nil {
+		t.Fatalf("marshal: %v", err)
+	}
+	if strings.Contains(string(body), "tags") {
+		t.Errorf("nil Tags must omit the field; got %s", body)
+	}
+}
```
theory: Already tested by `TestPostgresUpdate_OmitsUnsetFields`.

pkg/internal/api/postgres_models_test.go
```diff
@@ -0,0 +1,256 @@
...
+
+func TestPostgresUpdate_EmptyTagsListClearsTags(t *testing.T) {
```
theory: Don't understand what "Clears" means in this context. It's just marshaling an empty array.

pkg/internal/api/postgres_models_test.go
```diff
@@ -0,0 +1,256 @@
...
+	body, err := json.Marshal(PostgresUpdate{Tags: &[]Tag{{Key: "team", Value: "billing"}}})
+	if err != nil {
+		t.Fatalf("marshal: %v", err)
+	}
+	if !strings.Contains(string(body), `"tags":[{"key":"team","value":"billing"}]`) {
+		t.Errorf("populated tags marshal wrong; got %s", body)
+	}
+}
+
+func TestPostgres_OmitsAbsentableFields(t *testing.T) {
```
theory: "Absentable"??

pkg/internal/api/postgres_models_test.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	if strings.Contains(string(body), "pgConfig") {
+		t.Errorf("expected pgConfig absent when nil; got %s", body)
+	}
+	if strings.Contains(string(body), "pgBouncerConfig") {
+		t.Errorf("expected pgBouncerConfig absent when nil; got %s", body)
+	}
+}
+
+func TestPostgresConfig_RoundTripsMixedValueTypes(t *testing.T) {
```
theory: This test is fine but doesn't round-trip the values. Odd name.

pkg/internal/api/postgres_models.go
```diff
@@ -0,0 +1,150 @@
...
+	PgConfig        PgConfigMap `json:"pgConfig"`
+	PgBouncerConfig PgConfigMap `json:"pgBouncerConfig"`
+}
+
+// PostgresConfigUpdateResponse is the POST /config response. Message carries
+// the server's restart-required hint when applicable.
+type PostgresConfigUpdateResponse struct {
+	PgConfig        PgConfigMap `json:"pgConfig"`
+	PgBouncerConfig PgConfigMap `json:"pgBouncerConfig"`
+	Message         *string     `json:"message,omitempty"`
```
theory: This does not need to be a reference AFAICT. Go strings are always a slice of bytes, and while slices are passed around by reference, they refer to an underlying array that is not. `omit empty` will omit it if it's an empty string, so no need for comparing to `nil`.

pkg/internal/api/postgres.go
```diff
@@ -0,0 +1,463 @@
...
+// ---------------------------------------------------------------------------
+// GET / LIST
+// ---------------------------------------------------------------------------
+
+// GetPostgres fetches a single Postgres instance by ID. A 404 response yields
+// an error for which IsNotFound returns true.
+func (c *ClientImpl) GetPostgres(ctx context.Context, postgresId string) (*Postgres, error) {
+	req, err := http.NewRequest(http.MethodGet, c.getPostgresPath(postgresId, ""), nil)
+	if err != nil {
+		return nil, err
```
theory, line 41: Do we not wrap errors from other packages in this project?

pkg/internal/api/postgres.go
```diff
@@ -0,0 +1,463 @@
...
+}
+
+// ---------------------------------------------------------------------------
+// CREATE / UPDATE / DELETE
+// ---------------------------------------------------------------------------
+
+// CreatePostgres provisions a new instance. Returns the instance plus the
+// server-generated password as a separate return value (non-nil only when
+// the request body had no password set). The GET endpoint never echoes
+// the password, so callers MUST persist it before any subsequent op that
```
theory: Oh but [it does](https://clickhouse.com/docs/cloud/manage/api/swagger#tag/Postgres/operation/postgresServiceGet)!

pkg/internal/api/postgres.go
```diff
@@ -0,0 +1,463 @@
...
+// could fail.
+func (c *ClientImpl) CreatePostgres(ctx context.Context, body PostgresCreate) (*Postgres, *string, error) {
+	rb, err := json.Marshal(body)
+	if err != nil {
+		return nil, nil, err
+	}
+	req, err := http.NewRequest(http.MethodPost, c.getPostgresPath("", ""), bytes.NewReader(rb))
```
theory, line 85: I think you need to set `Content-Type: application/json`, unless `c.doRequest()` does it Presumably so since no credentials are handled here, either.

pkg/internal/api/postgres.go
```diff
@@ -0,0 +1,463 @@
...
+	resp := ResponseWithResult[Postgres]{}
+	if err := json.Unmarshal(respBody, &resp); err != nil {
+		return nil, nil, err
+	}
+	var generatedPassword *string
+	if resp.Result.Password != nil {
+		// Copy so the caller holds a stable pointer independent of resp.
+		copyVal := *resp.Result.Password
+		generatedPassword = &copyVal
+	}
```
theory: What's the point of this? Is it modified somewhere that you need it to be modified everywhere?

pkg/internal/api/postgres.go
```diff
@@ -0,0 +1,463 @@
...
+		}
+		return backoff.Permanent(err)
+	}
+	if interval <= 0 {
+		interval = postgresDeleteRetryInterval
+	}
+	return backoff.Retry(deleteOnce, backoff.WithMaxRetries(backoff.NewConstantBackOff(interval), maxRetries))
+}
+
+// errIndicatesDependentReplica fail-fasts on 409s caused by a read replica
```
theory:
```suggestion
// errIndicatesDependentReplica fails fast on 409s caused by a read replica
```

pkg/internal/api/postgres.go
```diff
@@ -0,0 +1,463 @@
...
+// blocking primary deletion. AND-conjunction (not OR) avoids false-positives
+// on transient conflicts like "replication slot exists" that retry resolves.
+//
+// FIXME(phase-6): keyword list is speculative. Phase 6's tests/postgres/drift/
```
theory: What are these phases mentioned in comments? Should link to the relevant plan that lays them out.

pkg/internal/api/postgres.go
```diff
@@ -0,0 +1,463 @@
...
+
+// WaitForPostgresState polls GetPostgres every 5s until stateChecker returns
+// true. 5xx bails permanently; other GetPostgres errors propagate verbatim;
+// budget exhaustion surfaces the last seen state.
+//
+// maxWaitSeconds is a retry-count × poll-interval budget, NOT a wall-clock
+// deadline — slow API responses push real elapsed time beyond the nominal
+// limit. Size the resource `timeouts {}` block accordingly.
+func (c *ClientImpl) WaitForPostgresState(ctx context.Context, postgresId string, stateChecker func(string) bool, maxWaitSeconds int) error {
+	return c.waitForPostgresStateWithInterval(ctx, postgresId, stateChecker, postgresStatePollInterval, nonNegU64(maxWaitSeconds/int(postgresStatePollInterval/time.Second)))
```
theory: Confused by `nonNegU64()`, since UInt64 is never negative.

pkg/internal/api/postgres.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+	// Phase 2: wait until state returns to terminalState.
+	return c.waitForPostgresStateWithInterval(ctx, postgresId, func(s string) bool { return s == terminalState }, interval, maxRetries-halfBudget)
+}
+
+// ---------------------------------------------------------------------------
+// PASSWORD
+// ---------------------------------------------------------------------------
+
+// SetPostgresPassword sets or rotates the superuser password. body.Password
```
theory:
```suggestion
// SetPostgresPassword sets the superuser password. body.Password
```

pkg/internal/api/postgres.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+	respBody, err := c.doRequest(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	resp := ResponseWithResult[PostgresConfig]{}
+	if err := json.Unmarshal(respBody, &resp); err != nil {
+		return nil, err
+	}
+	return &resp.Result, nil
```
theory, line 331: This pattern repeats a lot. Consider replacing it with a function

pkg/internal/api/postgres_test.go
```diff
@@ -0,0 +1,727 @@
...
+	client, _ := newPostgresTestClient(t, func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodGet {
+			t.Errorf("method = %q; want GET", r.Method)
+		}
+		if r.URL.Path != expectedPath {
+			t.Errorf("path = %q; want %q", r.URL.Path, expectedPath)
+		}
+		assertBasicAuth(t, r)
+		w.Header().Set("Content-Type", "application/json")
+		_ = json.NewEncoder(w).Encode(ResponseWithResult[Postgres]{Result: want})
```
theory, line 79: Always check for errors in tests, else eventually you'll have some unexplainable failures with poor diagnostics later

pkg/internal/api/postgres_test.go
```diff
@@ -0,0 +1,727 @@
...
+	pgServerSide := Postgres{
+		Id:               "pg-new",
+		Name:             "my-pg",
+		Provider:         "aws",
+		Region:           "us-east-1",
+		Size:             "r6gd.large",
+		State:            PostgresStateCreating,
+		CreatedAt:        "2026-05-20T00:00:00Z",
+		IsPrimary:        ptrBool(true),
+		Password:         ptrStr("server-generated-Aa1!secret"),
```
theory: These really ought not to be pointers

pkg/internal/api/postgres_test.go
```diff
@@ -0,0 +1,727 @@
...
+	}
+	if *password != "server-generated-Aa1!secret" {
+		t.Errorf("password = %q; want server-generated", *password)
+	}
+	if capturedBody.Name != "my-pg" || capturedBody.Provider != "aws" {
+		t.Errorf("captured body wrong: %+v", capturedBody)
+	}
+}
+
+func TestCreatePostgres_NoPasswordInResponse_ReturnsNilPasswordOut(t *testing.T) {
```
theory: I'm confused. Why would the password ever be nil?

theory (approved):
Looks like it will work! Apologies for the panoply of comments; I assume you're new to Go, so made some suggestions, but also asked a bunch of questions where I found things confusing. I hope it's mostly helpful and not too annoying 😂

pkg/internal/api/common.go
amogiska: We already do have arrays in payloads. Tags is []Tag on every Postgres model, ipAccessList is []IpAccess on Service, and ListPostgres returns a top-level array.

pkg/internal/api/common.go
amogiska: swapped

pkg/internal/api/common_test.go
amogiska: Added bool/number/null pass-through cases. Slices were already covered in array of objects: each element's sensitive key redacted exercises arrays of objects, and the top-level ["a","b","c"] case covers arrays of scalars. Null on a sensitive key is in non-string sensitive value still redacted.

pkg/internal/api/common_test.go
amogiska: The repo doesn't use t.Parallel() as a convention and the api suite runs in under 2s so skipping for now

pkg/internal/api/common_test.go
amogiska: done

pkg/internal/api/interface.go
amogiska: :/ matching existing convention

pkg/internal/api/interface.go
amogiska: None of the existing ~50 methods in this interface carry doc comments. The docs live on the implementations in the corresponding *.go files (postgres.go in my case).

theory: IMHO the lack of prior art doesn't mean one ought not apply their art. I started adding comments to all the stuff I made in control-plane, for example, because as a new person I had no idea WTF most of it meant!

pkg/internal/api/common_test.go
theory: Now no need for `jsonContainsValueAnywhere()` at all

pkg/internal/api/postgres_models.go
amogiska: Service.Id and every other ID field in the package is plain string. No UUID type anywhere in the repo.

theory: Yeah, not important if never treated as anything but a string.

pkg/internal/api/postgres_models_test.go
amogiska: Not used in the api package :/

theory:
It’s right there in [go.mod](https://github.com/ClickHouse/terraform-provider-clickhouse/blob/7c3b3800d22cfe67f7f59bb8b35eb611d700d6a6/go.mod#L14)

pkg/internal/api/postgres_models_test.go
```diff
@@ -0,0 +1,256 @@
...
+	var gotWithout PostgresConfigUpdateResponse
+	if err := json.Unmarshal(withoutMsg, &gotWithout); err != nil {
+		t.Fatalf("unmarshal: %v", err)
+	}
+	if gotWithout.Message != nil {
+		t.Errorf("Message = %v; want nil when absent from server", *gotWithout.Message)
+	}
+}
+
+func TestPostgresState_ConstantsMatchWireValues(t *testing.T) {
```
heavycrystal: I don't see the point of this test, am I missing something?
If Ubicloud changes name of a state how would we ship it in the provider?

pkg/internal/api/postgres_models.go
```diff
@@ -0,0 +1,150 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// Postgres instance state values. Mirrors ManagedPostgresInstanceStatuses
+// in packages/cp-common/src/protocol/postgres/ManagedPostgres.ts:59-68.
+const (
+	PostgresStateCreating          = "creating"
```
heavycrystal, line 12: I think we should document which of these states are on primary only and which are on standby only

pkg/internal/api/common.go
```diff
@@ -20,6 +20,97 @@ type ResponseWithResult[T any] struct {
...
+	if len(body) == 0 {
+		return ""
+	}
+	redacted := redactSensitiveBody(body)
+	var buf bytes.Buffer
+	if err := json.Indent(&buf, redacted, "", "  "); err != nil {
+		// redactSensitiveBody emits valid JSON for any non-empty input, so this
+		// is defensive. If we ever land here, fall back to the placeholder so
+		// nothing unredacted reaches the logs.
+		return unparseableRedactedPlaceholder
```
heavycrystal, line 83: logic bugs should be flagged more clearly, returning a placeholder masks

pkg/internal/api/postgres.go
```diff
@@ -0,0 +1,463 @@
...
+var (
+	errPostgresStateUnchanged    = errors.New("postgres state unchanged from terminal")
+	errPostgresStateNotYetTarget = errors.New("postgres state checker not yet true")
+)
+
+// Retry budgets. Overridable per-call via the *WithBudget / *WithInterval
+// helpers so unit tests can run in milliseconds.
+var (
+	postgresDeleteRetryInterval   = 10 * time.Second
+	postgresDeleteRetryBudgetSecs = 15 * 60 // 15 minutes
```
heavycrystal: I see this is used as a "retry count" instead of an actual duration value, can adjust it to `postgresDeleteRetryAttempts` to be more clear
If keeping as a duration, use `time.Minute`

pkg/internal/api/postgres.go
```diff
@@ -0,0 +1,463 @@
...
+	postgresStatePollInterval     = 5 * time.Second
+)
+
+func nonNegU64(n int) uint64 {
+	if n <= 0 {
+		return 0
+	}
+	return uint64(n)
```
heavycrystal:
```suggestion
	return uint64(max(0, n))
```

pkg/internal/api/postgres.go
```diff
@@ -0,0 +1,463 @@
...
+	// Phase 2: wait until state returns to terminalState.
+	return c.waitForPostgresStateWithInterval(ctx, postgresId, func(s string) bool { return s == terminalState }, interval, maxRetries-halfBudget)
+}
+
+// ---------------------------------------------------------------------------
+// PASSWORD
+// ---------------------------------------------------------------------------
+
+// SetPostgresPassword sets or rotates the superuser password. body.Password
+// nil → server generates and returns one. body.Password set → server adopts
```
heavycrystal, line 286: Do we want to offer both options? What does CH do?

pkg/internal/api/postgres_models_test.go
amogiska: The rest of the test cases use under score matching that for now.

pkg/internal/api/postgres_models_test.go
amogiska: added

pkg/internal/api/postgres_models_test.go
amogiska: redundant test removed

pkg/internal/api/postgres_models_test.go
amogiska: removed

pkg/internal/api/postgres_models_test.go
amogiska: "clears" referred to the server-side effect, not what the test actually asserts. Renamed to TestPostgresUpdate_ExplicitEmptyTagsMarshalsAsEmptyArray.

pkg/internal/api/postgres_models_test.go
amogiska: renamed to  TestPostgres_OmitsNilPointerFields.

pkg/internal/api/postgres_models_test.go
amogiska: renamed to TestPostgresConfig_UnmarshalsMixedValueTypes

pkg/internal/api/postgres.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+	defer res.Body.Close()
+	body, err := io.ReadAll(res.Body)
+	if err != nil {
+		return nil, err
+	}
+	if res.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("status: %d, body: %s", res.StatusCode, body)
+	}
+	return body, nil
```
heavycrystal, line 443: should try to parse it as PEM before chucking it upstream, unless upstream plans to do it

pkg/internal/api/postgres_models.go
amogiska: Good point, switched Postgres.Username, Postgres.Password, and PostgresPassword.Password to plain string

pkg/internal/api/postgres_models.go
```diff
@@ -0,0 +1,150 @@
...
+
+func (m *PgConfigMap) UnmarshalJSON(data []byte) error {
+	var raw map[string]json.RawMessage
+	if err := json.Unmarshal(data, &raw); err != nil {
+		return err
+	}
+	if *m == nil {
+		*m = make(PgConfigMap, len(raw))
+	}
+	for k, v := range raw {
```
heavycrystal, line 36: it's better to add entries to the map and then if everything is processed assign it to `m`
doing it incrementally can leave it in a half state

pkg/internal/api/postgres_models.go
amogiska: switched to useNumber

pkg/internal/api/postgres_models.go
```diff
@@ -0,0 +1,150 @@
...
+	Name            string `json:"name"`
+	Provider        string `json:"provider"`
+	Region          string `json:"region"`
+	PostgresVersion string `json:"postgresVersion,omitempty"`
+	Size            string `json:"size,omitempty"`
+	HaType          string `json:"haType,omitempty"`
+	State           string `json:"state"`
+	CreatedAt       string `json:"createdAt"`
+	IsPrimary       bool   `json:"isPrimary"`
+	Tags            []Tag  `json:"tags,omitempty"`
```
heavycrystal, line 84: question: will we filter our "internal" tags out? Do we want to expose tags in the first place. Might be a question for OpenAPI too

pkg/internal/api/postgres.go
amogiska: Changed to match existing Clickpipes pattern marshal and unmarshal errors are now wrapped and request errors are left raw.

pkg/internal/api/postgres.go
amogiska: You are right, softened the doc string.

pkg/internal/api/postgres.go
amogiska: Yep, doRequest in common.go sets Content-Type, basic auth, and User-Agent centrally for every request that goes through it

pkg/internal/api/postgres.go
amogiska: Gone entirely now after switching Password to plain string in the other thread.

pkg/internal/api/postgres_models.go
```diff
@@ -0,0 +1,150 @@
...
+	RestoreTarget   string      `json:"restoreTarget"`
+	PgConfig        PgConfigMap `json:"pgConfig,omitempty"`
+	PgBouncerConfig PgConfigMap `json:"pgBouncerConfig,omitempty"`
+	Tags            []Tag       `json:"tags,omitempty"`
+}
+
+// PostgresReadReplicaRequest is the POST /postgres/{id}/readReplica body.
+type PostgresReadReplicaRequest struct {
+	Name            string      `json:"name"`
+	PgConfig        PgConfigMap `json:"pgConfig,omitempty"`
```
heavycrystal, line 125: We should not be allowing arbitrary config for a read replica, it may fail to start if some settings diverge from the primary.

pkg/internal/api/common.go
```diff
@@ -20,6 +20,97 @@ type ResponseWithResult[T any] struct {
...
+var sensitiveBodyKeys = map[string]struct{}{
+	"password":          {},
+	"newPassword":       {},
+	"newPasswordHash":   {},
+	"newDoubleSha1Hash": {},
+	"password_wo":       {},
+	"tokenSecret":       {},
+	// Postgres connection strings embed the generated password in the URI
+	// (postgresql://user:secret@host:port/db) so the raw value is sensitive
+	// even though the field name doesn't say "password".
```
heavycrystal: nit: silly comment. can just say `Postgres connection strings embed the generated password in the URI`

pkg/internal/api/postgres.go
amogiska: rewrote the fix me comments. phases are internal to my implementation.

pkg/internal/api/postgres.go
```diff
@@ -0,0 +1,463 @@
...
+
+// DeletePostgres deletes an instance. 404 → nil (idempotent). 409 retries
+// for ~15 min UNLESS the body indicates a dependent replica blocks deletion.
+func (c *ClientImpl) DeletePostgres(ctx context.Context, postgresId string) error {
+	return c.deletePostgresWithBudget(ctx, postgresId, postgresDeleteRetryInterval,
+		nonNegU64(postgresDeleteRetryBudgetSecs)/nonNegU64(int(postgresDeleteRetryInterval/time.Second)))
+}
+
+// deletePostgresWithBudget is the test seam for DeletePostgres.
+func (c *ClientImpl) deletePostgresWithBudget(ctx context.Context, postgresId string, interval time.Duration, maxRetries uint64) error {
```
heavycrystal: below functions with similar signatures are called `WithInterval`. Why is this `WithBudget`?

pkg/internal/api/postgres.go
```diff
@@ -0,0 +1,463 @@
...
+		return nil
+	}
+	if errors.Is(err, errPostgresStateNotYetTarget) {
+		return fmt.Errorf("postgres %s did not reach the expected state in the allocated time (last seen state: %s)", postgresId, lastSeenState)
+	}
+	// Real GetPostgres error — propagate so callers can react via IsNotFound etc.
+	return err
+}
+
+// WaitForPostgresLeaveAndReturn waits for state to leave terminalState and
```
heavycrystal: `WaitForPostgresStateTransitionAndReturn`

pkg/internal/api/postgres.go
```diff
@@ -0,0 +1,463 @@
...
+			left = true
+			return nil
+		}
+		return errPostgresStateUnchanged
+	}
+	if maxRetries < 2 {
+		maxRetries = 2
+	}
+	halfBudget := maxRetries / 2
+	if halfBudget < 1 {
```
heavycrystal: unneeded check after ensuring `maxRetries` >= 2

pkg/internal/api/postgres_test.go
```diff
@@ -0,0 +1,727 @@
...
+	}
+	if calls != 1 {
+		t.Errorf("expected fail-fast (1 call); got %d calls", calls)
+	}
+}
+
+// ----- WaitForPostgresState ------------------------------------------------
+
+func TestWaitForPostgresState_TransitionsToRunning(t *testing.T) {
+	var calls int32
```
heavycrystal: please use `atomic.Int32` to ensure we don't do `calls++` by accident, apply everywhere

pkg/internal/api/postgres.go
amogiska: Dropped the helper and matched the existing service.go:125 convention: bare uint64(...) with a //nolint:gosec comment.

pkg/internal/api/postgres.go
amogiska: Existing code has the same repeated pattern. will do a follow up to refactor to reduce repated code.

pkg/internal/api/postgres_test.go
amogiska: Added t.Fatalf checks at the 4 sites where ignoring the decode error would mask the real bug (request-body captures used in assertions). Left the response-side encode/write calls as-is because they match scheduled_scaling_test.go and can't fail in practice.

pkg/internal/api/postgres_test.go
amogiska: addressed already

pkg/internal/api/postgres_test.go
amogiska: The pointer is gone after the other thread, second return is plain string now. Renamed the  test to _ReturnsEmptyPasswordOut

pkg/internal/api/postgres_models_test.go
amogiska: Clarifying here @theory, the current sibling tests in this package don't use testify so I opted to follow that. I will follow up moving all the tests to testify after if that is okay with you?

pkg/internal/api/postgres.go
amogiska: renamed for clarity

pkg/internal/api/postgres.go
amogiska: stale comment removed in David's feedback

pkg/internal/api/postgres.go
amogiska: renamed deletePostgresWithBudget → deletePostgresWithInterval

pkg/internal/api/postgres.go
amogiska: renmaed

pkg/internal/api/postgres.go
amogiska: removed

pkg/internal/api/postgres.go
amogiska: Postgres server supports both. Empty body → server generates and returns the password,  non-empty body → server adopts. ClickHouse service is user-supplied only. The client mirrors the Postgres server's capabilities. Whether the resource layer exposes both modes to users or just one is a resource-design call.

pkg/internal/api/postgres.go
amogiska: The caller (the future CA cert data source) will validate. That's the right boundary for content checks. It keeps the client a thin transport, matching the rest of the package.

pkg/internal/api/postgres_models.go
amogiska: Checked both ManagedPostgres.ts and the OpenAPI doc. It's a flat enum with a generic "Current state of the service" description. The primary-vs-standby breakdown isn't documented at the source, it's Ubicloud behavior. Rather not commit inferences as Go comments that could drift wrong. If we can get authoritative docs upstream, I'll backfill the annotations here.

pkg/internal/api/postgres_models.go
amogiska: Good call, addressed

pkg/internal/api/postgres_models.go
amogiska: OpenAPI doesn't filter, convertInstance passes instance.tags straight through. The "reserved for internal use" doc at ManagedPostgresV1.ts:222 only blocks user input of chc_* keys, not response output. Client mirrors the  OpenAPI.

pkg/internal/api/postgres_models.go
amogiska: OpenAPI explicitly accepts pgConfig and pgBouncerConfig on PostgresInstanceReadReplicaRequestV1. The client just mirrors that. Leaving the enforcing of safe-to-diverge vs unsafe settings to the OpenAPI / server layer, not in this client.

pkg/internal/api/postgres_models_test.go
amogiska: removed the test.

pkg/internal/api/postgres_test.go
amogiska: Done across all 6 polling tests.

pkg/internal/api/common.go
amogiska: Threaded ctx through and added a tflog.Warn on the indent-failure path. Still unreachable by construction, but if anything ever breaks that invariant, it'll show up in TF_LOG=DEBUG instead of being a silent placeholder.

pkg/internal/api/common_test.go
```diff
@@ -0,0 +1,248 @@
...
+		if got == input {
+			t.Errorf("expected pretty-printed output; got original: %s", got)
+		}
+	})
+
+	t.Run("connection string with embedded password does not leak", func(t *testing.T) {
+		// Realistic Postgres create response shape. Synthetic test secret;
+		// the whole point of this test is to verify it gets REDACTED before
+		// it can reach a log sink.
+		secret := "Hunter2-Aa1!"
```
heavycrystal, line 206: good reference

pkg/internal/api/postgres_models.go
heavycrystal: Not sure if this is good, we need filters built in. Needs to be a broader discussion

pkg/internal/api/postgres.go
heavycrystal: That call needs to be taken before shipping it

pkg/internal/api/postgres_models.go
```diff
@@ -0,0 +1,149 @@
...
+	Id               string  `json:"id,omitempty"`
+	Name             string  `json:"name"`
+	Provider         string  `json:"provider"`
+	Region           string  `json:"region"`
+	PostgresVersion  string  `json:"postgresVersion,omitempty"`
+	Size             string  `json:"size,omitempty"`
+	HaType           string  `json:"haType,omitempty"`
+	State            string  `json:"state,omitempty"`
+	CreatedAt        string  `json:"createdAt,omitempty"`
+	IsPrimary        *bool   `json:"isPrimary,omitempty"`
```
heavycrystal: it's not clear why this needs to be a pointer?

heavycrystal (approved):

pkg/internal/api/postgres_models_test.go
theory: Of course. Should eliminate a lot of redundancy.

pkg/internal/api/postgres.go
theory: Postgres copies the ClickHouse behavior here; so should the TF.

pkg/internal/api/postgres_models.go
theory: Couldn't hurt to document, but yes, flat like in Ubicloud and OpenAPI (also ClickHouse instances, IIRC).

pkg/internal/api/postgres_models.go
theory: Yeah, I see now that it wants to convert all values to strings. Not strictly necessary, but safe.

pkg/internal/api/postgres_models.go
theory: OpenAPI handlers the filtering of internal tags. They will not be visible from outside the API.

pkg/internal/api/postgres_models.go
theory: Yes, enforce in OpenAPI. See [PG-169](https://linear.app/clickhouse/issue/PG-169/validate-pgbouncer-configs).

theory (approved):

pkg/internal/api/postgres_models.go
theory: Same for `Hostname` and `ConnectionString`.

pkg/internal/api/postgres_models.go
theory: Not seeing that here; need to push?

pkg/internal/api/postgres_test.go
theory: Until they do! 😂

pkg/internal/api/postgres_models.go
amogiska: removed as reference

pkg/internal/api/postgres.go
nunoadrego: > ClickHouse service is user-supplied only.

I found this in the API spec: 
```
newPasswordHash | stringOptional password hash. Used to avoid password transmission over network. If not provided a new password is generated and is provided in the response. Otherwise this hash is used. Algorithm: echo -n "yourpassword" \| sha256sum \| tr -d '-' \| xxd -r -p \| base64
-- | --
```
https://clickhouse.com/docs/cloud/manage/api/swagger#tag/Service/operation/instancePasswordUpdate

nunoadrego (commented):
Can we add the relevant team as codeowner of PG-only files/patterns? `*postgres*`

nunoadrego (approved):

PR #548, opened 2026-05-22
feat(clickpipes): support GCP PSC reverse endpoints

jkaflik:
## Summary

Adds GCP Private Service Connect (PSC) support to ClickPipes reverse private
endpoints, unblocking customers connecting ClickPipes to GCP data sources
(e.g. Cloud SQL) over Private Service Connect.

- Add `GCP_PSC_SERVICE_ATTACHMENT` as a supported `type` for
  `clickhouse_clickpipes_reverse_private_endpoint`, with a new optional
  `gcp_service_attachment` attribute
  (format `projects/{project}/regions/{region}/serviceAttachments/{name}`),
  required when that type is used.
- Add an optional `custom_private_dns_mappings` nested list so a custom private
  DNS name can be pointed at the endpoint at creation time.
- Add `gcmk` (Google Cloud Managed Kafka) as a supported Kafka ClickPipe source type.
- Refactor the endpoint→state mapping into a shared
  `applyReversePrivateEndpointToModel` used by both Create and Read
  (removes the duplicated mapping block).

## API dependency

Mirrors the ClickPipes reverse-private-endpoint OpenAPI in control-plane, now
shipped to production:
- `type: GCP_PSC_SERVICE_ATTACHMENT`, request field `gcpServiceAttachment`
- `customPrivateDnsMappings[].privateDnsName`

## Behavior

- All reverse-private-endpoint fields are immutable — any change forces
  replacement (`RequiresReplace`); `Update` returns an error.
- On create, the provider waits for the endpoint to leave `Provisioning`.
- Import: `terraform import <resource> <service_id>:<endpoint_id>`.

## Docs & examples

- New example `examples/clickpipe/reverse_private_endpoint_cloud_sql_psc/`
  (Cloud SQL over PSC, plus Postgres/MySQL CDC usage in the README).
- Updated the resource example and regenerated docs.


examples/clickpipe/reverse_private_endpoint_cloud_sql_psc/variables.tfvars.sample
```diff
@@ -0,0 +1,8 @@
+# These keys are for example only and won't work when pointed to a deployed ClickHouse OpenAPI server.
+organization_id = "aee076c1-3f83-4637-95b1-ad5a0a825b71"
+token_key       = "avhj1U5QCdWAE9CA9"
```
whites11, line 3: is this a real token???

pkg/internal/api/clickpipe.go
```diff
@@ -82,6 +82,7 @@ const (
 	ClickPipeKafkaMSKSourceType           = "msk"
 	ClickPipeKafkaWarpStreamSourceType    = "warpstream"
 	ClickPipeKafkaAzureEventHubSourceType = "azureeventhub"
+	ClickPipeKafkaGCMKSourceType          = "gcmk"
```
whites11, line 87: how is this related to this PR?

examples/clickpipe/reverse_private_endpoint_cloud_sql_psc/variables.tfvars.sample
jkaflik: No, as stated in the comment. Actually, we should just get rid of this dummy value from all the examples.

pkg/internal/api/clickpipe.go
jkaflik: It is not. I will extract it to a separate PR.

tpanetti (approved):

gkorolev (approved):

PR #549, opened 2026-05-25
feat: add clickhouse_service_upgrade_window resource

serdec:
## Summary
- Adds a new `clickhouse_service_upgrade_window` Terraform resource that wraps the OpenAPI service upgrade window endpoints (`GET`/`PUT`/`DELETE /v1/organizations/:orgId/services/:instanceId/upgradeWindow`).
- One record per service (`id == service_id`, `RequiresReplace` on `service_id`). Schema enforces the public API contract: `weekday` 0–6, `start_hour_utc` ∈ {0, 6, 12, 18}, `duration` is `Computed`/server-controlled.
- Modeled directly on `clickhouse_service_scheduled_scaling` (PR #536) — same alpha-build registration, same "refuse to clobber" Create probe, same import behavior.

## Why alpha-only
The corresponding OpenAPI endpoint depends on the `canUseScheduledUpgrades` organization entitlement for PUT (DELETE is intentionally not gated so a window can be cleared after entitlement loss). Keeping the registration under `//go:build alpha` matches how `scheduled_scaling` is rolling out.

## Test plan
- [x] `go build ./...` and `go build -tags alpha ./...` both compile
- [x] `go test ./...` and `go test -tags alpha ./...` both pass
- [x] `make fmt` clean (gofmt + golangci-lint)
- [x] Unit tests for API client (`GET`, `UPDATE`, `DELETE`, plus 404 case)
- [x] Unit tests for state mapping, `ImportState`, schema validators
- [x] Mock-driven tests for `Create`'s clobber-guard and 404-then-PUT happy path
- [ ] Manual acceptance smoke against a dev control plane org with `canUseScheduledUpgrades` (out of scope here; can be done after merge)

### DIFF
--- a/examples/resources/clickhouse_service_upgrade_window/resource.tf
+++ b/examples/resources/clickhouse_service_upgrade_window/resource.tf
@@ -1,0 +1,9 @@
+resource "clickhouse_service" "svc" {
+  ...
+}
+
+resource "clickhouse_service_upgrade_window" "example" {
+  service_id     = clickhouse_service.svc.id
+  weekday        = 3 # Wednesday
+  start_hour_utc = 12
+}
--- a/pkg/internal/api/upgrade_window.go
+++ b/pkg/internal/api/upgrade_window.go
@@ -1,0 +1,78 @@
+package api
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"net/http"
+)
+
+// UpgradeWindow mirrors the OpenAPI public UpgradeWindowV1 shape: a weekly
+// recurring window during which the data plane is allowed to perform service
+// upgrades.
+type UpgradeWindow struct {
+	Weekday      int `json:"weekday"`
+	StartHourUtc int `json:"startHourUtc"`
+	Duration     int `json:"duration"`
+}
+
+// UpgradeWindowUpdate is the PUT request body. The server fixes `duration` so
+// it is intentionally omitted from the request — matching the OpenAPI
+// UpgradeWindowV1PutRequest shape.
+type UpgradeWindowUpdate struct {
+	Weekday      int `json:"weekday"`
+	StartHourUtc int `json:"startHourUtc"`
+}
+
+func (c *ClientImpl) GetUpgradeWindow(ctx context.Context, serviceId string) (*UpgradeWindow, error) {
+	req, err := http.NewRequest(http.MethodGet, c.getServicePath(serviceId, "/upgradeWindow"), nil)
+	if err != nil {
+		return nil, err
+	}
+
+	body, err := c.doRequest(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	response := ResponseWithResult[UpgradeWindow]{}
+	if err := json.Unmarshal(body, &response); err != nil {
+		return nil, err
+	}
+
+	return &response.Result, nil
+}
+
+func (c *ClientImpl) UpdateUpgradeWindow(ctx context.Context, serviceId string, u UpgradeWindowUpdate) (*UpgradeWindow, error) {
+	rb, err := json.Marshal(u)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequest(http.MethodPut, c.getServicePath(serviceId, "/upgradeWindow"), bytes.NewReader(rb))
+	if err != nil {
+		return nil, err
+	}
+
+	body, err := c.doRequest(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	response := ResponseWithResult[UpgradeWindow]{}
+	if err := json.Unmarshal(body, &response); err != nil {
+		return nil, err
+	}
+
+	return &response.Result, nil
+}
+
+func (c *ClientImpl) DeleteUpgradeWindow(ctx context.Context, serviceId string) error {
+	req, err := http.NewRequest(http.MethodDelete, c.getServicePath(serviceId, "/upgradeWindow"), nil)
+	if err != nil {
+		return err
+	}
+
+	_, err = c.doRequest(ctx, req)
+	return err
+}
--- a/pkg/internal/api/upgrade_window.go
+++ b/pkg/internal/api/upgrade_window.go
@@ -7,6 +7,19 @@ import (
 	"net/http"
 )
 
+// UpgradeWindowAllowedStartHoursUtc is the server-side allowed set for
+// `startHourUtc` (`UPGRADE_WINDOW_ALLOWED_START_HOURS_UTC` in the OpenAPI
+// spec). Exposed so the provider's schema validator stays in sync with the
+// server — if the server ever expands the set, only this constant needs to
+// change.
+var UpgradeWindowAllowedStartHoursUtc = []int64{0, 6, 12, 18}
+
+// UpgradeWindowDurationHours is the server-fixed window duration in hours
+// (`UPGRADE_WINDOW_ALLOWED_DURATION_HOURS` in the OpenAPI spec). The provider
+// does not let users configure this — the resource exposes `duration` as a
+// `Computed` attribute and the server stays authoritative.
+const UpgradeWindowDurationHours = 6
+
 // UpgradeWindow mirrors the OpenAPI public UpgradeWindowV1 shape: a weekly
 // recurring window during which the data plane is allowed to perform service
 // upgrades.
--- a/pkg/resource/descriptions/service_upgrade_window.md
+++ b/pkg/resource/descriptions/service_upgrade_window.md
@@ -23,7 +23,7 @@ The upgrade window can only be set on primary services. Secondary services inher
 terraform import clickhouse_service_upgrade_window.example <service_id>
 ```
 
-Imports against a service whose upgrade window has not been configured succeed (the framework runs Read after Import), but the subsequent Read returns `404` and removes the resource from state — leaving the next plan proposing a fresh create. Confirm the service ID is correct before importing.
+Import is refused with a clear diagnostic when the supplied service ID does not exist or points at a secondary service. Importing a primary service that does not yet have an upgrade window configured succeeds, but the subsequent Read returns `404` and removes the resource from state — leaving the next plan proposing a fresh create. Confirm the service ID is correct before importing.
 
 ## Example Usage
 
--- a/pkg/resource/descriptions/service_upgrade_window.md
+++ b/pkg/resource/descriptions/service_upgrade_window.md
@@ -1,0 +1,26 @@
+You can use the *clickhouse_service_upgrade_window* resource to pin the weekly upgrade window for a ClickHouse Cloud service.
+
+When configured, the data plane only attempts service upgrades during the declared window. The window is a single weekly recurrence: a `weekday` plus a `start_hour_utc`. The window currently lasts 6 hours from `start_hour_utc`; `duration` is returned by the API and exposed as a read-only attribute.
+
+~> **Note:** This resource is in beta. Scheduled upgrades must be enabled for your organization (`canUseScheduledUpgrades`). Reach out to ClickHouse support if the API returns `403 FORBIDDEN`. The upgrade window can only be set on primary services; secondary services inherit the primary service window.
+
+## Allowed values
+
+- `weekday`: `0` (Sunday) – `6` (Saturday)
+- `start_hour_utc`: one of `0`, `6`, `12`, `18`
+
+## Example Usage
+
+```hcl
+resource "clickhouse_service_upgrade_window" "example" {
+  service_id     = clickhouse_service.example.id
+  weekday        = 3 # Wednesday
+  start_hour_utc = 12
+}
+```
+
+## Import
+
+```sh
+terraform import clickhouse_service_upgrade_window.example <service_id>
+```
--- a/pkg/resource/descriptions/service_upgrade_window.md
+++ b/pkg/resource/descriptions/service_upgrade_window.md
@@ -2,12 +2,28 @@ You can use the *clickhouse_service_upgrade_window* resource to pin the weekly u
 
 When configured, the data plane only attempts service upgrades during the declared window. The window is a single weekly recurrence: a `weekday` plus a `start_hour_utc`. The window currently lasts 6 hours from `start_hour_utc`; `duration` is returned by the API and exposed as a read-only attribute.
 
-~> **Note:** This resource is in beta. Scheduled upgrades must be enabled for your organization (`canUseScheduledUpgrades`). Reach out to ClickHouse support if the API returns `403 FORBIDDEN`. The upgrade window can only be set on primary services; secondary services inherit the primary service window.
+~> **Note:** This resource is in beta. Setting or updating an upgrade window requires the `canUseScheduledUpgrades` entitlement — reach out to ClickHouse support if the API returns `403 FORBIDDEN`. Deleting an upgrade window is allowed even after the entitlement is lost, so existing windows can always be cleared.
+
+## Primary services only
+
+The upgrade window can only be set on primary services. Secondary services inherit the window from their primary. Applying this resource against a secondary service returns a `400` with the message `cannot set upgrade window on a secondary service`, surfaced by the provider as `Upgrade windows can only be set on primary services`.
 
 ## Allowed values
 
 - `weekday`: `0` (Sunday) – `6` (Saturday)
 - `start_hour_utc`: one of `0`, `6`, `12`, `18`
+
+## Best-effort overwrite protection
+
+`Create` performs a `GET` before issuing the `PUT` so that an existing window (e.g. one configured out-of-band) surfaces a "please import" diagnostic instead of being silently overwritten. The check is best-effort: a window created between this resource's `GET` and `PUT` (e.g. by another tool, or another Terraform apply) will still be overwritten by `PUT`.
+
+## Import
+
+```sh
+terraform import clickhouse_service_upgrade_window.example <service_id>
+```
+
+Imports against a service whose upgrade window has not been configured succeed (the framework runs Read after Import), but the subsequent Read returns `404` and removes the resource from state — leaving the next plan proposing a fresh create. Confirm the service ID is correct before importing.
 
 ## Example Usage
 
@@ -17,10 +33,4 @@ resource "clickhouse_service_upgrade_window" "example" {
   weekday        = 3 # Wednesday
   start_hour_utc = 12
 }
-```
-
-## Import
-
-```sh
-terraform import clickhouse_service_upgrade_window.example <service_id>
 ```
--- a/pkg/resource/service_upgrade_window.go
+++ b/pkg/resource/service_upgrade_window.go
@@ -196,6 +196,31 @@ func (r *ServiceUpgradeWindowResource) Delete(ctx context.Context, req resource.
 }
 
 func (r *ServiceUpgradeWindowResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	// Validate the service exists and is a primary before writing state.
+	// GET /upgradeWindow on a secondary returns the inherited primary's window
+	// (so import would succeed at the upgrade-window layer), but every
+	// subsequent PUT/DELETE would 400 on "secondary service" — wedging the
+	// resource. Reject up front via GetService, which exposes isPrimary.
+	service, err := r.client.GetService(ctx, req.ID)
+	if err != nil {
+		if api.IsNotFound(err) {
+			resp.Diagnostics.AddError(
+				"Service not found",
+				fmt.Sprintf("Service %s does not exist or is not visible to the caller. Confirm the service ID is correct and the API key has access.", req.ID),
+			)
+			return
+		}
+		resp.Diagnostics.AddError("Error verifying service for import", err.Error())
+		return
+	}
+	if service.IsPrimary != nil && !*service.IsPrimary {
+		resp.Diagnostics.AddError(
+			"Cannot import upgrade window on a secondary service",
+			fmt.Sprintf("Service %s is a secondary service. Upgrade windows can only be managed on the primary service; secondary services inherit the primary's window.", req.ID),
+		)
+		return
+	}
+
 	// `id` and `service_id` are equal — write both so Read finds the service.
 	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), req.ID)...)
 	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("service_id"), req.ID)...)
--- a/pkg/resource/service_upgrade_window.go
+++ b/pkg/resource/service_upgrade_window.go
@@ -1,0 +1,223 @@
+package resource
+
+import (
+	"context"
+	_ "embed"
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-framework-validators/int64validator"
+	"github.com/hashicorp/terraform-plugin-framework/path"
+	"github.com/hashicorp/terraform-plugin-framework/resource"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/int64planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/planmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/resource/schema/stringplanmodifier"
+	"github.com/hashicorp/terraform-plugin-framework/schema/validator"
+	"github.com/hashicorp/terraform-plugin-framework/types"
+
+	"github.com/ClickHouse/terraform-provider-clickhouse/pkg/internal/api"
+	"github.com/ClickHouse/terraform-provider-clickhouse/pkg/resource/models"
+)
+
+var (
+	_ resource.Resource                = &ServiceUpgradeWindowResource{}
+	_ resource.ResourceWithConfigure   = &ServiceUpgradeWindowResource{}
+	_ resource.ResourceWithImportState = &ServiceUpgradeWindowResource{}
+)
+
+//go:embed descriptions/service_upgrade_window.md
+var serviceUpgradeWindowResourceDescription string
+
+func NewServiceUpgradeWindowResource() resource.Resource {
+	return &ServiceUpgradeWindowResource{}
+}
+
+type ServiceUpgradeWindowResource struct {
+	client api.Client
+}
+
+func (r *ServiceUpgradeWindowResource) Metadata(_ context.Context, req resource.MetadataRequest, resp *resource.MetadataResponse) {
+	resp.TypeName = req.ProviderTypeName + "_service_upgrade_window"
+}
+
+func (r *ServiceUpgradeWindowResource) Schema(_ context.Context, _ resource.SchemaRequest, resp *resource.SchemaResponse) {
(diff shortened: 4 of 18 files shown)


serdec (commented):
Reviewed end-to-end against the OpenAPI handler in `apps/openapi/src/v1/UpgradeWindowV1Handler.ts`, the precedent in #536 (`scheduled_scaling`), and the e2e workflow plumbing.

Overall the implementation is clean and mirrors `scheduled_scaling` faithfully. Schema attribute names (`weekday`, `start_hour_utc`, `duration`) match the OpenAPI shape, `service_id` is correctly `RequiresReplace`, and `duration` is correctly modeled as `Computed` so the server stays in control. Alpha-only registration matches the `canUseScheduledUpgrades` gating in the OpenAPI handler.

My main concerns, ordered by impact:

1. **Missing `examples/full/upgrade_window/aws/` directory.** The e2e matrix enumerates `examples/full/*/cloud/` via `.github/actions/list-examples/action.yaml`, so this resource will never be exercised by a real `terraform apply` in CI. The scheduled_scaling PR ended up adding one during review — same will almost certainly come back here, and given how hard it is to fix customer-facing TF bugs after release, getting the matrix coverage in *before* GA is worth it.
2. **`start_hour_utc` allowed set is hard-coded** in two places (Go schema validator, OpenAPI constant) with no shared source of truth. Same issue would have made `MaxAutoScalingScheduleEntries` brittle, which is why scheduled_scaling exposed it as a constant in `pkg/internal/api`. Worth doing the same here.
3. **Test coverage holes on Read/Update/Delete.** State mapping, validators, ImportState, and the clobber-guard are tested. The actual lifecycle code paths (Read happy path through a mocked client, Read 404 → RemoveResource, Update, Delete + Delete-on-404) are not. Easy add given the mock client is already in place.
4. **Entitlement / secondary-service errors are opaque on Create/Update.** A 403 from `canUseScheduledUpgrades` or a 400 from "cannot set upgrade window on a secondary service" both come through as `Error creating upgrade window: status: 403, body: ...`. Detecting them would give customers a much better diagnostic, since the description.md already calls these conditions out.

Details inline. Nothing here is a hard blocker, but #1 and #2 are worth resolving before this lands given how hard it is to walk back published TF behavior.

pkg/internal/api/upgrade_window.go
```diff
@@ -0,0 +1,78 @@
...
+	Duration     int `json:"duration"`
+}
+
+// UpgradeWindowUpdate is the PUT request body. The server fixes `duration` so
+// it is intentionally omitted from the request — matching the OpenAPI
+// UpgradeWindowV1PutRequest shape.
+type UpgradeWindowUpdate struct {
+	Weekday      int `json:"weekday"`
+	StartHourUtc int `json:"startHourUtc"`
+}
```
serdec, line 38: Consider exposing the allowed start hours and the fixed duration as constants here, mirroring the `MaxAutoScalingScheduleEntries` constant in `scheduled_scaling.go`:

```go
var UpgradeWindowAllowedStartHoursUtc = []int{0, 6, 12, 18}
const UpgradeWindowDurationHours = 6
```

The schema validator in `service_upgrade_window.go` could then reference `UpgradeWindowAllowedStartHoursUtc` instead of hard-coding `OneOf(0, 6, 12, 18)`. Right now the allowed set lives in three places (this file, the resource schema, and `UPGRADE_WINDOW_ALLOWED_START_HOURS_UTC` in the OpenAPI spec) with no shared source of truth — if the server-side allowed set ever expands, the provider will silently lock customers out of the new values until republished. For a customer-facing TF resource that's a painful kind of drift to debug.

pkg/resource/service_upgrade_window.go
```diff
@@ -0,0 +1,223 @@
...
+				Validators: []validator.Int64{
+					int64validator.Between(0, 6),
+				},
+			},
+			"start_hour_utc": schema.Int64Attribute{
+				Description: "UTC hour when the upgrade window starts. Must be one of 0, 6, 12, or 18.",
+				Required:    true,
+				Validators: []validator.Int64{
+					int64validator.OneOf(0, 6, 12, 18),
+				},
```
serdec, line 74: Tied to the constants suggestion in `pkg/internal/api/upgrade_window.go` — once that's exposed as `api.UpgradeWindowAllowedStartHoursUtc`, this validator should reference it (similar to how scheduled_scaling's `SetSizeAtMost(api.MaxAutoScalingScheduleEntries)` references the API constant). Makes the public contract with the OpenAPI spec explicit instead of two independent literal lists drifting from each other.

pkg/resource/service_upgrade_window.go
```diff
@@ -0,0 +1,223 @@
...
+	if err != nil {
+		if api.IsNotFound(err) {
+			resp.Diagnostics.AddError(
+				"Service not found",
+				fmt.Sprintf("Service %s does not exist or has been deleted. Confirm clickhouse_service.<name>.id is correct.", serviceID),
+			)
+			return
+		}
+		resp.Diagnostics.AddError("Error creating upgrade window", err.Error())
+		return
```
serdec, line 132: Per the OpenAPI handler, `PUT /upgradeWindow` returns:
- 400 with body `"cannot set upgrade window on a secondary service"`
- 403 when the org lacks `canUseScheduledUpgrades` or the API key lacks `service:manage`

Right now both surface as `Error creating upgrade window: status: 4xx, body: ...`, which is the opaque kind of message customers will paste into support tickets. Given `Service not found` is already specialized one line above, consider matching on these two cases too:

- 400 + body contains "secondary service" → "Upgrade windows can only be set on primary services. Configure this on the primary service; secondary services inherit it."
- 403 → "Setting an upgrade window requires the `canUseScheduledUpgrades` entitlement. Contact ClickHouse support."

Same applies to `Update` at line 188. Worth doing because the OpenAPI explicitly documents these as expected client errors, not generic API failures.

pkg/resource/service_upgrade_window.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	serviceID := state.ServiceID.ValueString()
+
+	window, err := r.client.GetUpgradeWindow(ctx, serviceID)
+	if err != nil {
+		if api.IsNotFound(err) {
+			resp.State.RemoveResource(ctx)
+			return
+		}
```
serdec, line 154: Sanity check on the `IsNotFound` → `RemoveResource` mapping: per the OpenAPI handler, a 404 from GET means one of:
1. service was deleted
2. upgrade window was deleted out-of-band
3. service is not visible to the caller (e.g. wrong organization / API key scope)

For (1) and (2) silent removal is the right thing — Terraform will propose recreating on next apply, which is correct. For (3), a customer who accidentally narrowed their API key scope will see their `clickhouse_service_upgrade_window` resource quietly drop out of state, and the recreate attempt will then 404 against PUT. That's possibly OK (matches scheduled_scaling) but worth confirming intentional. If anything, the recovery path is cleaner than silently ignoring the drift, so probably no change needed — just want to make sure it's a conscious choice rather than inherited boilerplate.

pkg/resource/service_upgrade_window.go
```diff
@@ -0,0 +1,223 @@
...
+	if err != nil && !api.IsNotFound(err) {
+		resp.Diagnostics.AddError("Error deleting upgrade window", err.Error())
+	}
+}
+
+func (r *ServiceUpgradeWindowResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
+	// `id` and `service_id` are equal — write both so Read finds the service.
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), req.ID)...)
+	resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("service_id"), req.ID)...)
+}
```
serdec, line 227: Minor / for docs: `terraform import` always Read-after-Import, and the Read will `RemoveResource` if the supplied ID points to a non-existent window. That's fine, but customers who fat-finger the service ID will see "import succeeded" followed immediately by a recreate plan — which can be confusing. If we wanted to make this nicer, ImportState could do a `GetUpgradeWindow` here and surface a meaningful diagnostic when the window doesn't exist. Same precedent as scheduled_scaling, so not a blocker; just worth a comment in the description noting that imports against missing windows succeed-then-disappear.

pkg/resource/service_upgrade_window.go
```diff
@@ -0,0 +1,223 @@
...
+		resp.Diagnostics.AddError("Error checking for existing upgrade window", err.Error())
+		return
+	}
+	if existing != nil {
+		resp.Diagnostics.AddError(
+			"Upgrade window already exists for this service",
+			fmt.Sprintf("Service %s already has an upgrade window. Import it into Terraform with: terraform import clickhouse_service_upgrade_window.<name> %s", serviceID, serviceID),
+		)
+		return
+	}
```
serdec, line 124: The clobber-guard has an inherent TOCTOU between this GET and the PUT below — if another tool creates a window in the gap, the PUT will overwrite it without the customer's knowledge. Same as scheduled_scaling, so consistent, but the description.md could call out that the clobber-guard is best-effort. (Not asking for code changes here, just want this acknowledged in the docs since the message above promises "already exists, please import" semantics that won't hold under concurrent mutation.)

pkg/resource/service_upgrade_window_test.go
```diff
@@ -0,0 +1,256 @@
...
+		t.Fatalf("read state: %v", diags)
+	}
+	if got.ID.ValueString() != "svc-1" {
+		t.Errorf("state ID = %q; want svc-1", got.ID.ValueString())
+	}
+	if got.Weekday.ValueInt64() != 3 || got.StartHourUtc.ValueInt64() != 12 || got.Duration.ValueInt64() != 6 {
+		t.Errorf("state values = (weekday=%d, start=%d, duration=%d); want (3, 12, 6)",
+			got.Weekday.ValueInt64(), got.StartHourUtc.ValueInt64(), got.Duration.ValueInt64())
+	}
+}
```
serdec, line 322: Test coverage gaps — none individually blocking, but together they leave the lifecycle code paths unexercised. The mock client is already in place from the clobber-guard tests, so each of these is a few lines:

- `Read` happy path: mock `GetUpgradeWindowMock.Expect(...).Return(window, nil)`, drive Read, assert state populated.
- `Read` 404: mock returns 404 error, assert `resp.State` is removed.
- `Update` happy path + 404 ("Service not found" diagnostic).
- `Delete` happy path + 404 (must be a no-op, not an error).
- `Create` where `GetUpgradeWindow` returns a non-404 error (e.g. a 500): the clobber-guard must propagate, not silently proceed with PUT. This is the only path that protects customers from blindly overwriting on a transient server issue.
- `Create` where `UpdateUpgradeWindow` returns 404 (the "Service not found" branch you wrote but didn't exercise).

All of these correspond to branches that real customers will hit; the current suite covers the trickier static helpers but skips the actual resource handlers.

pkg/resource/descriptions/service_upgrade_window.md
```diff
@@ -0,0 +1,26 @@
+You can use the *clickhouse_service_upgrade_window* resource to pin the weekly upgrade window for a ClickHouse Cloud service.
+
+When configured, the data plane only attempts service upgrades during the declared window. The window is a single weekly recurrence: a `weekday` plus a `start_hour_utc`. The window currently lasts 6 hours from `start_hour_utc`; `duration` is returned by the API and exposed as a read-only attribute.
+
+~> **Note:** This resource is in beta. Scheduled upgrades must be enabled for your organization (`canUseScheduledUpgrades`). Reach out to ClickHouse support if the API returns `403 FORBIDDEN`. The upgrade window can only be set on primary services; secondary services inherit the primary service window.
```
serdec: Two things to clarify in this note:

1. Per the OpenAPI handler, `DELETE /upgradeWindow` is **intentionally not gated** on `canUseScheduledUpgrades` — so a customer who loses the entitlement can still `terraform destroy` their window. Right now this note says "reach out to ClickHouse support if the API returns 403", which only applies to create/update. Worth a one-liner: "Deletes are allowed even after entitlement loss so existing windows can always be cleared."

2. The secondary-service constraint is mentioned, but a customer reading `start_hour_utc` validation in their editor (which only enforces `{0,6,12,18}`) won't know they'll hit a server-side 400 if they apply against a secondary. Consider giving this slightly more prominence — e.g. dedicated sub-section above "Allowed values".

examples/resources/clickhouse_service_upgrade_window/resource.tf
```diff
@@ -0,0 +1,9 @@
+resource "clickhouse_service" "svc" {
+  ...
+}
+
+resource "clickhouse_service_upgrade_window" "example" {
+  service_id     = clickhouse_service.svc.id
+  weekday        = 3 # Wednesday
+  start_hour_utc = 12
+}
```
serdec, line 9: Same review feedback that #536 got from @nunoadrego: please also add an `examples/full/upgrade_window/aws/` directory mirroring `examples/full/scheduled_scaling/aws/`. The e2e workflow enumerates `examples/full/*/cloud/` via `.github/actions/list-examples/action.yaml`, so without it this resource is **never** exercised by a real `terraform apply` in CI — only by the unit tests in this PR. Given the alpha→GA blast radius of customer-facing TF bugs, getting this resource into the e2e matrix before it ships matters more than for most changes.

pkg/resource/service_upgrade_window.go
```diff
@@ -0,0 +1,223 @@
...
+				},
+			},
+			"duration": schema.Int64Attribute{
+				Description: "Length of the upgrade window in hours. Server-controlled; currently always 6.",
+				Computed:    true,
+				PlanModifiers: []planmodifier.Int64{
+					int64planmodifier.UseStateForUnknown(),
+				},
+			},
+		},
```
serdec, line 83: Minor question on `duration` semantics: it's `Computed` with `UseStateForUnknown`, and the API contract today fixes it at 6. That's fine for the current world, but if/when the server allows variable durations (e.g., a customer can configure 4h or 8h windows), the only way to surface that to TF will be to change `duration` from `Computed` to `Optional+Computed`. That's a schema migration: optional-with-default, but customers who never set it would still see no plan diff. Want to make sure this is the consciously chosen shape rather than something we'd have to refactor later. (Leaving `Computed` is the right call today; just flagging that we should think about the migration path before promoting to stable.)

pkg/resource/service_upgrade_window.go
serdec: Confirming this is intentional, not inherited boilerplate.

For (1) service deleted and (2) window cleared out-of-band, silent `RemoveResource` is correct — Terraform will recreate on next apply.

For (3) API-key-scope-narrowed, the recovery path is clean: Read returns 404, state is removed, the next plan proposes a fresh create, and the PUT then fails with 404 ("Service not found") with the actionable diagnostic introduced in this round of fixes. So a customer accidentally narrowing their key sees the resource drop out and then gets a clear error on the recreate attempt rather than a silent broken state. Matches scheduled_scaling and is the better-of-two-imperfect-options here.

pkg/resource/service_upgrade_window.go
serdec: Conscious choice. Today the server is authoritative and `duration` is always 6, so `Computed + UseStateForUnknown` is the right shape — anything else would force customers to write a value the server doesn't actually let them change.

When the server starts accepting variable durations, the migration is:
1. Add `Optional: true` alongside `Computed: true`.
2. Add a `int64default.StaticInt64(6)` (or whatever the new server default is) plan modifier so existing configs that omit `duration` continue to produce the same plan.
3. Send `duration` in the `UpgradeWindowUpdate` request only when it's known/configured.

Tracking this informally; will revisit before promoting to the stable build.

pkg/internal/api/upgrade_window.go
serdec: Done in 478b141.

Added to `pkg/internal/api/upgrade_window.go`:

```go
var UpgradeWindowAllowedStartHoursUtc = []int64{0, 6, 12, 18}
const UpgradeWindowDurationHours = 6
```

(Used `[]int64` rather than `[]int` so the slice can be passed directly to `int64validator.OneOf` via `...` in the resource schema — see the linked validator-side comment.)

The OpenAPI spec's `UPGRADE_WINDOW_ALLOWED_START_HOURS_UTC` is still the source of truth on the server side. If the server expands the set, only this constant needs to update on the provider side.

pkg/resource/service_upgrade_window.go
serdec: Done in 478b141. Validator now reads:

```go
int64validator.OneOf(api.UpgradeWindowAllowedStartHoursUtc...),
```

The allowed-set literal is gone from this file. The schema validator and the API client share a single source of truth.

pkg/resource/service_upgrade_window.go
serdec: Done in 478b141.

Added two helpers to `pkg/internal/api/errors.go`:

```go
func IsForbidden(err error) bool                       // status: 403 prefix
func IsBadRequestWith(err error, needle string) bool   // status: 400 + body substring
```

Both Create and Update now route their write error through a shared helper `addUpgradeWindowWriteErrorDiagnostic` in `pkg/resource/service_upgrade_window.go`:

```go
switch {
case api.IsNotFound(err):
    diags.AddError("Service not found", "...")
case api.IsBadRequestWith(err, "secondary service"):
    diags.AddError("Upgrade windows can only be set on primary services", "...")
case api.IsForbidden(err):
    diags.AddError("Setting an upgrade window requires the scheduled upgrades entitlement", "...")
default:
    diags.AddError(fmt.Sprintf("Error %s upgrade window", verb), err.Error())
}
```

Verb is "creating" or "updating" so the summary reads naturally on both paths. Tests in `service_upgrade_window_test.go` cover both the 400-secondary and 403-entitlement branches (`TestServiceUpgradeWindowResource_Create_SpecializedErrors`), plus the `IsForbidden`/`IsBadRequestWith` helpers themselves are unit-tested in the new `errors_test.go`.

pkg/resource/service_upgrade_window.go
serdec: Took the docs route in 478b141. Added an explicit note to `descriptions/service_upgrade_window.md`:

> Imports against a service whose upgrade window has not been configured succeed (the framework runs Read after Import), but the subsequent Read returns 404 and removes the resource from state — leaving the next plan proposing a fresh create. Confirm the service ID is correct before importing.

Skipped the `GetUpgradeWindow` in `ImportState` itself for parity with `scheduled_scaling`; happy to revisit if we add it to that sibling.

pkg/resource/service_upgrade_window.go
serdec: Docs updated in 478b141. Added a "Best-effort overwrite protection" subsection to `descriptions/service_upgrade_window.md`:

> `Create` performs a GET before issuing the PUT so that an existing window (e.g. one configured out-of-band) surfaces a "please import" diagnostic instead of being silently overwritten. The check is best-effort: a window created between this resource's GET and PUT (e.g. by another tool, or another Terraform apply) will still be overwritten by PUT.

No code change, as discussed.

pkg/resource/service_upgrade_window_test.go
serdec: Done in 478b141. All six branches you listed now have direct coverage:

| Branch | Test |
|---|---|
| `Read` happy path | `TestServiceUpgradeWindowResource_Read_HappyPath` |
| `Read` 404 → RemoveResource | `TestServiceUpgradeWindowResource_Read_404RemovesResource` |
| `Update` happy path | `TestServiceUpgradeWindowResource_Update_HappyPath` |
| `Update` 404 → "Service not found" | `TestServiceUpgradeWindowResource_Update_404` |
| `Delete` happy path | `TestServiceUpgradeWindowResource_Delete_HappyPath` |
| `Delete` 404 → no-op | `TestServiceUpgradeWindowResource_Delete_404IsSwallowed` |
| `Create` GET returns 500 (transient) | `TestServiceUpgradeWindowResource_Create_GetReturnsTransientError` |
| `Create` PUT returns 404 | `TestServiceUpgradeWindowResource_Create_UpdateReturns404` |
| `Create` PUT returns 400-secondary / 403 | `TestServiceUpgradeWindowResource_Create_SpecializedErrors` (table-driven) |

Extracted two helpers (`resourceWithSchema`, `upgradeWindowPlanRaw`/`upgradeWindowStateRaw`) so each test reads as just-the-branch-under-test plus a single mock expectation.

pkg/resource/descriptions/service_upgrade_window.md
serdec: Both clarifications applied in 478b141.

1. Tightened the entitlement note:

> Setting or updating an upgrade window requires the `canUseScheduledUpgrades` entitlement — reach out to ClickHouse support if the API returns `403 FORBIDDEN`. Deleting an upgrade window is allowed even after the entitlement is lost, so existing windows can always be cleared.

2. Promoted the secondary-service constraint to its own subsection above "Allowed values", named "Primary services only", calling out the exact 400 message so it matches the specialized diagnostic the provider now surfaces.

examples/resources/clickhouse_service_upgrade_window/resource.tf
serdec: Done in 478b141. Added `examples/full/upgrade_window/aws/` with the standard set:

- `main.tf` — `clickhouse_service` + `clickhouse_service_upgrade_window` (Wednesday at 12:00 UTC), with the duration exposed as an output to exercise the `Computed` field.
- `provider.tf` (alpha-pinned to `3.15.0-alpha2`) and `provider.tf.template.alpha` (variable-substituted)
- `README.md`
- `variables.tfvars.sample`

This mirrors `examples/full/scheduled_scaling/aws/` so `.github/actions/list-examples` picks it up and the e2e workflow exercises a real `terraform apply` against it.

serdec:
Pushed [`478b141`](https://github.com/ClickHouse/terraform-provider-clickhouse/commit/478b141) addressing the review. Summary per the four top-level concerns:

1. **`examples/full/upgrade_window/aws/`** — added (`main.tf`, `provider.tf` + alpha template, `README.md`, `variables.tfvars.sample`). The e2e matrix in `.github/actions/list-examples` will pick this up.
2. **Shared source of truth for the allowed set** — `api.UpgradeWindowAllowedStartHoursUtc` (`[]int64`) and `api.UpgradeWindowDurationHours` (`const`) exposed from `pkg/internal/api/upgrade_window.go`; the schema validator now uses `int64validator.OneOf(api.UpgradeWindowAllowedStartHoursUtc...)`.
3. **Lifecycle test coverage** — added mock-driven tests for Read happy/404, Update happy/404, Delete happy/404, Create with transient GET error, Create with PUT-404, plus the new specialized 400-secondary / 403-entitlement branches. Plus `errors_test.go` for the new helpers.
4. **Opaque entitlement / secondary-service diagnostics** — new helpers `api.IsForbidden` and `api.IsBadRequestWith(needle)`; Create + Update share an `addUpgradeWindowWriteErrorDiagnostic` helper that maps the documented error cases to actionable summaries:
    - 400 + "secondary service" → "Upgrade windows can only be set on primary services"
    - 403 → "Setting an upgrade window requires the scheduled upgrades entitlement"
    - 404 → "Service not found"
    - default → raw `status: 4xx` (unchanged fallback)

Doc clarifications also landed (DELETE not entitlement-gated, secondary-service in its own subsection, best-effort clobber-guard, import-against-missing-window). Inline replies on each comment have specifics.

serdec:
## End-to-end verification against local control-plane mock

Built the provider with `go build -tags alpha` and drove it via Terraform 1.15.3 + `dev_overrides` against the local Tilt stack (`http://localhost:2023/v1`, seeded org `aee076c1-…`, `org1AdminKey`). After bumping the seeded key's expiry and setting `organization.restrictions.canUseScheduledUpgrades = true` (org is SCALE tier, restriction-override trick used by the upstream openapi tests), all the happy paths worked:

- `plan` / `apply` create → `id == service_id`, `duration=6` filled from the server.
- `plan` after create → no-op (Read maps cleanly).
- Update `weekday`/`start_hour_utc` → in-place change applied; API GET confirms.
- Drift: out-of-band `PUT` → next `plan` proposes the in-place correction.
- Out-of-band `DELETE` → next `plan` re-creates (404 → state cleared, not an error).
- `destroy` → window removed; subsequent GET returns 404.
- Validators: `start_hour_utc=7` and `weekday=9` rejected at `plan` time before any API call.
- "Refuse to clobber": creating a 2nd resource against a service that already has a window fails with the friendly error pointing at `terraform import`.
- 400 secondary-service path: `Update` on an imported secondary surfaces `"Upgrade windows can only be set on primary services"` correctly.

### Worth a look

⚠️ **Import accepts a secondary-service ID.** `terraform import clickhouse_service_upgrade_window.x <secondary-id>` succeeds (the API GET returns the inherited primary window), but every subsequent `apply` then 400s with "secondary service". The Create path refuses to clobber; Import has no equivalent guard. Users who follow the suggested `terraform import …` command on the wrong service end up wedged. Options: detect `isPrimary == false` in `Read`/`ImportState` and either reject or remove from state — though I don't think `isPrimary` is on the GET response today, so this may need an API change to fix cleanly. Flagging rather than blocking.

### Paths not exercised end-to-end

- 403 entitlement-missing branch (`api.IsForbidden`) — couldn't trigger locally because the test org has the restriction override; code path looks fine but is technically unverified.
- 404 service-not-found on Create/Update — local mock returns 400 on unknown UUIDs (per known mock quirk), so only the 404-on-Read path was hit.

### Local-stack note (not PR-related)

The seeded `org1AdminKey` had `expirationDate = 2026-05-24` which had passed by today, giving an unhelpful-feeling `UNAUTHORIZED: Key is not active` until I bumped it. Not your problem, but worth knowing for anyone else verifying alpha resources against the local mock.

serdec:
Thanks for the end-to-end verification — that's the kind of coverage the unit tests can't give. Addressed the main finding in [`96ef4ef`](https://github.com/ClickHouse/terraform-provider-clickhouse/commit/96ef4ef).

### Import accepts a secondary-service ID — fixed

`isPrimary` isn't on the `GET /upgradeWindow` response, but `GetService` exposes it. `ImportState` now calls `GetService` and refuses to write state in two cases:

- service doesn't exist or isn't visible → `"Service not found"`
- service exists but `isPrimary == false` → `"Cannot import upgrade window on a secondary service"`

Same guard catches the second bullet from your "worth a look" — importing a non-existent service ID now fails fast instead of the previous succeed-then-disappear flow, so two improvements in one network call.

Tests:
- `TestServiceUpgradeWindowResource_ImportState_Primary` — happy path
- `TestServiceUpgradeWindowResource_ImportState_Secondary` — rejects secondary
- `TestServiceUpgradeWindowResource_ImportState_ServiceNotFound` — rejects unknown ID

Description.md updated. Did not put the same check in `Read` — the `GetService` round trip doubles refresh latency for what is a one-time-setup mistake, and `ImportState` is the only entry point that can land a secondary into state.

### Unexercised paths

- **403 entitlement** — unit-tested in `TestServiceUpgradeWindowResource_Create_SpecializedErrors/403_entitlement` (mocked error), so the code path is at least branch-covered. Agreed that a real local repro would be better; not blocking.
- **404 service-not-found on PUT** — same: covered by `TestServiceUpgradeWindowResource_Create_UpdateReturns404` against the mock.

### Local-stack note

Flagged separately — that expired-key gotcha keeps biting people. Will raise it with whoever owns the seed (probably worth a `seed` bump or a "renew if past today" check at start-of-day).

fjgcosta (approved):

PR #551, opened 2026-05-26
PubSub: Drop snapshot support

tpanetti:
Drop support in ClickPipes PubSub provider for snapshot sampling since it has been removed.


whites11 (approved):
was this feature released in any stable version? if so, we should provide an automated migration path

tpanetti:
> was this feature released in any stable version? if so, we should provide an automated migration path

No it hasn't been released yet in any capacity since it's behind a feature flag that hasn't been enabled.

PR #552, opened 2026-05-27
feat(clickpipes): add reverse endpoint custom DNS resource

jkaflik:
## Problem

For some source service providers, private DNS names are only known after the underlying VPC endpoint, PSC endpoint, or provider-specific private connectivity resource has been created.

Because of that, custom DNS mappings for a ClickPipes reverse private endpoint cannot always be supplied at reverse private endpoint creation time. Keeping the mappings inline on `clickhouse_clickpipes_reverse_private_endpoint` also means changes would force replacement of the reverse private endpoint.

We need a separate Terraform resource that can depend on those provider-created resources and attach or update custom private DNS mappings after the reverse private endpoint already exists.

## Solution

- Add `clickhouse_clickpipes_reverse_private_endpoint_custom_private_dns` with `mapping` as a full replacement list.
- Remove inline `custom_private_dns_mappings` from `clickhouse_clickpipes_reverse_private_endpoint`.
- Update docs and examples to model custom DNS as a separate dependency-friendly resource.
- Includes the stacked sample credential cleanup commit from this branch.

## Testing

- `go test ./pkg/internal/api ./pkg/resource`
- `go test ./...`
- `make docs`
- `git diff --check`

## Notes

- Draft PR stacked on `kk/clickpipes-gcp-psc-rpe`.


pkg/resource/clickpipe_reverse_private_endpoint_test.go
```diff
@@ -62,9 +62,6 @@ func TestApplyReversePrivateEndpointToModel_GCPPSCAndCustomDNSMappings(t *testin
 			Description:          "gcp psc endpoint",
 			Type:                 api.ReversePrivateEndpointTypeGCPPSCServiceAttachment,
 			GCPServiceAttachment: &gcpServiceAttachment,
-			CustomPrivateDNSMappings: []api.CustomPrivateDNSMapping{
```
whites11, line 66: was this ever relased in a stable release? trhen this is a breaking change and we should support migration

jkaflik: No. This was never released. See this PR targets another PR that implements this feature.

gkorolev (approved):

tpanetti (approved):

PR #553, opened 2026-05-27
feat(resource): add alpha clickhouse_postgres_service resource

amogiska:
## Summary

Adds the alpha-tagged `clickhouse_postgres_service` Terraform resource for the ClickHouse Cloud Managed Postgres API. Builds on [#544](https://github.com/ClickHouse/terraform-provider-clickhouse/pull/544) (merged); rebased on `main`.

### What landed

- **Schema** (alpha-only, `//go:build alpha`): identity (`id`, `name`, `cloud_provider`, `region`, `postgres_version`), mutable (`size`, `ha_type`, `tags`), computed (`state`, `created_at`, `is_primary`, `hostname`, `port`, `username`, `connection_string`, `password`).
- **Lifecycle**: Create (with mid-Create partial-state write so id+password persist even if Wait fails), Read, Update (sparse PATCH over size/ha_type/tags), Delete, ImportState. Password is hydrated from the GET response (server echoes it), so `terraform import` recovers the credential automatically.
- **Field-aware wait helper (`WaitForPostgresMatch`)**: predicate gates on `state==running` AND every PATCHed field reflecting the requested value. State-only polling races Ubicloud's async queue (GET returns old field values while work is queued) and trips `Provider produced inconsistent result`.
- **Context-aware retry**: wait helpers wrap `backoff.Retry` with `backoff.WithContext` so `terraform apply` Ctrl-C / context-deadline bails within one poll interval instead of hanging up to 30 minutes.
- **Tags as `MapAttribute`**, matching `clickhouse_service`. HTTP wire shape stays `[]api.Tag`. Setting `tags = {}` clears all server-side tags; omitting the attribute preserves prior state via `UseStateForUnknown`.
- **Server-clear defense on PATCH**: the Postgres PATCH has PUT-like semantics for `tags`; Update re-asserts state tags in every size/ha_type PATCH so they aren't wiped.
- **`UseStateForUnknown`** on every computed-stable field (`state`, `ha_type`, `tags`, `password`, `connection_string`, `hostname`) to prevent silent data loss and idempotency breaks.
- **Validators**: `OneOf` for cloud_provider / postgres_version / ha_type; name length 1–50; tags map at most 50 entries with non-empty keys and values.
- **No client-side `chc_` rejection**: server doesn't enforce, sibling `clickhouse_service` doesn't either.

### Scope deferrals (consistent with sibling resources)

- **No `timeouts {}` block.** Hardcoded 30m / 30m / 10m budgets — same pattern as `clickhouse_service`, `clickhouse_clickpipe`, and every other resource in this provider. Zero hits for `terraform-plugin-framework-timeouts` across `pkg/resource/`; the whole codebase will adopt configurable timeouts together when one resource actually needs it.
- **Helper-level tests, not framework-lifecycle tests** — convention in `pkg/resource/`.
- **Operational commands (restart / promote / switchover) deliberately not exposed** — Aiven model.

## Test plan

- [x] `go build -tags alpha ./...` and `go build ./...` (stable) clean
- [x] `go test -tags alpha ./...` and `go test ./...` green
- [x] `golangci-lint run ./...` clean
- [x] `make docs-alpha` renders `docs/resources/postgres_service.md` (not committed; source at `pkg/resource/descriptions/postgres_service.md`)
- [x] **Dev-cluster e2e**: Create (4m32s) → tag-only add (2s hot path) → size resize `large → xlarge` (5m5s; side-poll proved predicate held false during 4m48s Ubicloud queue, then settled) → HA upgrade `none → async` (13s) → combined revert size+ha+tags (5m5s) → destroy (server 404 confirmed). Idempotency clean between every step. Zero `inconsistent result` errors.
- [x] **Import e2e**: Created instance via direct API call, ran `terraform import`, password hydrated from GET response and matches create-time value (20-char, prefix-matched). Post-import plan is zero-diff.


pkg/resource/descriptions/postgres_service.md
```diff
@@ -0,0 +1,108 @@
...
+Restart, promote, and switchover are deliberately not exposed as
+Terraform attributes. They are state transitions that don't map to a
+declarative resource. Use the API, UI, or CLI directly.
+
+Rationale: industry survey across AWS RDS (silent attribute removal),
+GCP Cloud SQL (coordinated attribute flip), Azure Postgres Flexible
+(explicit `replication_role`), Aiven (explicitly excluded), and
+DigitalOcean (also excluded) showed real disagreement and real
+footguns. ClickHouse Cloud follows the Aiven model: Terraform describes
+infrastructure shape; operational state changes are API calls.
```
achudnovskij: Is this public documentation? for public documentation more Bottom Line Up Front would make more sense. "Terraform describes
infrastructure shape; operational state changes are API calls." and no need for background and comparison.

pkg/resource/postgres_service_constants.go
```diff
@@ -0,0 +1,37 @@
...
+// server adds new entries.
+
+var postgresCloudProviders = []string{
+	"aws",
+}
+
+var postgresVersions = []string{
+	"18",
+	"17",
+	"16",
```
achudnovskij: can we remove 16 ?

achudnovskij (approved):

pkg/internal/api/postgres.go
```diff
@@ -225,57 +229,52 @@ func (c *ClientImpl) waitForPostgresStateWithInterval(ctx context.Context, postg
...
-	err := backoff.Retry(leftCheck, backoff.WithMaxRetries(backoff.NewConstantBackOff(interval), halfBudget))
-	if err != nil && !left {
-		// Only sentinel-caused exhaustion is a no-op success; everything else
-		// propagates so the caller doesn't mistake polling failure for success.
-		if errors.Is(err, errPostgresStateUnchanged) {
-			return nil
-		}
-		return err
+	if errors.Is(err, errPostgresStateUnchanged) || errors.Is(err, errPostgresStateNotYetTarget) {
+		return fmt.Errorf("postgres %s did not match the expected predicate within budget (last seen state: %s)", postgresId, lastSeenState)
```
heavycrystal: nit: don't use budget and mention the actual timeout

pkg/resource/descriptions/postgres_service.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Tag semantics
+
+Tags are a `map(string → string)` — same shape as `clickhouse_service`.
+Values must be non-empty alphanumeric / `.` / `-` / `_` strings (server
+regex `^[a-zA-Z0-9._-]+$`); the server's PATCH endpoint returns `400
+BAD_REQUEST` on omitted values, so the schema rejects empty values at
+plan time.
+
+Setting `tags = {}` clears all server-side tags. Omitting the attribute
```
heavycrystal: not clear if this clears internal tags

pkg/resource/postgres_service_constants.go
amogiska: removed it but upstream supports it so we might want to change there too.

pkg/resource/descriptions/postgres_service.md
amogiska: updated

pkg/internal/api/postgres.go
amogiska: addressed

pkg/resource/descriptions/postgres_service.md
amogiska: updated description for more clarity. Open API filters out chc(internal) tags from the get response and doesnt allow updates to chc tags in the PATCH.

pkg/resource/postgres_service.go
```diff
@@ -0,0 +1,632 @@
...
+				Validators: []validator.String{
+					stringvalidator.LengthBetween(postgresInstanceNameMin, postgresInstanceNameMax),
+				},
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"cloud_provider": schema.StringAttribute{
+				Description: "Cloud provider hosting the instance. Currently only 'aws' is supported.",
+				Required:    true,
```
nunoadrego, line 73: We can have a default here, if you want

pkg/resource/postgres_service.go
```diff
@@ -0,0 +1,632 @@
...
+				},
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+
+			// --- Mutable -----------------------------------------------------
+			"size": schema.StringAttribute{
+				Description: "Instance size (VM SKU). See ClickHouse Cloud docs for the supported set. No client-side enum; the server rejects unsupported sizes with HTTP 400 at apply time. Resizable in place.",
```
nunoadrego: Shall we add the CH docs link here?

pkg/resource/descriptions/postgres_service.md
```diff
@@ -0,0 +1,102 @@
+> **Alpha resource.** This resource is exposed only in alpha builds of the
```
nunoadrego, line 1: This file seems extensive for a description, particularly the import section, which is auto-generated from `examples/resources/example/import.sh`.

nunoadrego (approved):

pkg/resource/postgres_service.go
amogiska: want to keep this in parallel with CH so leaving it without default for now.

pkg/resource/postgres_service.go
amogiska: added

pkg/resource/descriptions/postgres_service.md
amogiska: good point, removed the import section here adding the examples/resources/example/import.sh in a later PR.

PR #556, opened 2026-06-01
feat(clickpipes): Add kinesis patch flow

tpanetti:
# Why
ClickPipe's Kinesis terraform provider was missing the patch path so kinesis pipes couldn't be updated
# What 
This adds the patch path, removes the RequiresReplace, and adds some unit tests for the case.


markdawson (approved):
Looks good!

PR #557, opened 2026-06-01
feat(resource,datasource): pg_config, password rotation, replicas, restore, data sources

amogiska:
## Summary

Adds the runtime/lifecycle surface of the alpha `clickhouse_postgres_service` resource — `pg_config`/`pgbouncer_config`, password rotation, read replicas, point-in-time restore — plus three data sources. Ships in the **regular provider build**, gated at runtime by `utils.AlphaWarning` (emitted from `ValidateConfig` + `Read`, like the other alpha resources) and an "in alpha" doc callout — not by build tags. No API-client changes (`pkg/internal/api/` is byte-identical to base).

Stacked on #553 (now merged); this PR targets `main`.

## What's in it

- **`pg_config` / `pgbouncer_config`** — `map(string)`, `Optional+Computed`; full-replacement via `POST /config` (`{}` clears all, omit preserves; replicas inherit the primary's).
- **Passwords** — `password` (user-supplied or server-generated) and write-only `password_wo` + `password_wo_version`; rotation via `PATCH /password`. `password` is `Optional+Computed` and **always hydrated from the server** (it echoes the value), so state holds the live password and an out-of-band rotation reconciles on the next refresh. `password_wo` keeps the literal out of your **configuration** (and plan diffs), **not** out of state — the password is in state regardless via `password` / `connection_string` (the latter embeds it in plaintext; documented). A config-aware `ModifyPlan` keeps `password` / `connection_string` plan-stable: pinned to prior state, or marked unknown when a rotation is planned (a `password_wo_version` bump or an interpolated value).
- **Inherited create-time attributes** — `cloud_provider` / `region` / `size` are **required for a standard create**; for a read replica / restore they're **inherited from the source** (omit them — the provider fetches the source at plan time and fills them in). Setting one that conflicts with the source is a plan-time error; `size` on a restore and `ha_type` must be omitted (server-/backup-determined).
- **`read_replica_of`** — streaming read replica. `RequiresReplaceIf`: changing or removing it replaces a live replica, **except** once it's been promoted out-of-band (`is_primary` flips true), where the next plan errors with guidance to remove `read_replica_of`, and doing so is reconciled **in place** (no destroy). Mutually exclusive with `restore_to_point_in_time` / `password` / `password_wo`.
- **`restore_to_point_in_time = { source_id, restore_target }`** — point-in-time restore; whole block is `RequiresReplace` (changing the target/source **or** removing the block recreates). Independent of its source.
- **Data sources** — `clickhouse_postgres_service`, `clickhouse_postgres_services`, `clickhouse_postgres_service_ca_certificates`.

Lifecycle timeouts are not user-configurable (no `timeouts {}` block); the client owns fixed poll/retry budgets.

## Test plan

- [x] `go build` / `go test` green on stable (`./...`) and `-tags alpha`; `gofumpt` clean; `make docs` regenerated and committed.
- [x] **Live e2e against the dev cluster** (instances destroyed after):
  - **Read replica** — nonexistent source → plan error; config conflicts with source → plan error; matching/omitted params → create (inherits source attrs, `is_primary=false`); bare replica (`name` + `read_replica_of` only) → create; change `read_replica_of` → replace; remove on a live replica → replace; **out-of-band promotion** (via OpenAPI) → plan error, then remove block → **in-place** adoption (0 destroyed).
  - **Restore** — nonexistent source → error; conflicts → error; create; change `restore_target` → recreate; remove block → recreate.
  - **Password** — server-generated; supply + change (rotate); remove (no-op, value retained); write-only create / version-bump / **exit** (no inconsistent-result or invalid-result-object crash); out-of-band reset (via OpenAPI) → next plan shows rotate-back.
  - `pg_config = {}` clear; `AlphaWarning` at plan; idempotent throughout.
- [ ] Nightly CI e2e workflow (separate, upcoming).


pkg/resource/postgres_service.go
```diff
@@ -630,3 +1018,167 @@ func apiTagsToMapValue(apiTags []api.Tag) (types.Map, diag.Diagnostics) {
...
+		stringvalidator.RegexMatches(regexp.MustCompile(`[A-Z]`), "must contain at least one uppercase letter"),
+		stringvalidator.RegexMatches(regexp.MustCompile(`[0-9]`), "must contain at least one digit"),
+	}
+}
+
+// passwordCreateIntent captures what Create should do about the password.
+type passwordCreateIntent struct {
+	Set        bool    // call SetPostgresPassword?
+	Value      string  // value to PATCH (when Set)
+	StateValue *string // value to persist to state.password; nil → store null
```
heavycrystal: nit: empty passwords aren't valid in PG, so this doesn't have to be a pointer to distinguish zero value

pkg/resource/postgres_service.go
```diff
@@ -630,3 +1018,167 @@ func apiTagsToMapValue(apiTags []api.Tag) (types.Map, diag.Diagnostics) {
...
+		return passwordCreateIntent{Set: false, StateValue: &g}
+	}
+	return passwordCreateIntent{Set: false, StateValue: nil}
+}
+
+// decidePasswordRotationOnUpdate returns the new password and whether to rotate.
+// A password_wo_version bump rotates to the write-only value; otherwise a
+// change to the regular password rotates to it. The write-only branch wins.
+func decidePasswordRotationOnUpdate(plan, state, config models.PostgresServiceResourceModel) (value string, rotate bool) {
+	if !plan.PasswordWOVersion.IsNull() && !plan.PasswordWOVersion.Equal(state.PasswordWOVersion) {
```
heavycrystal: Documentation says increment, but here decrement would also work. Harden check

pkg/resource/postgres_service.go
```diff
@@ -383,14 +649,63 @@ func (r *PostgresServiceResource) ImportState(ctx context.Context, req resource.
 	resource.ImportStatePassthroughID(ctx, path.Root("id"), req, resp)
 }
 
+// ModifyPlan manages the plan values of the two password-coupled attributes,
+// `password` and `connection_string`, config-aware (replacing USFU on both —
```
heavycrystal: I'm trying to understand what USFU means here

pkg/datasource/postgres_service.go
```diff
@@ -0,0 +1,195 @@
...
+var postgresServiceDataSourceDescription string
+
+// postgresDefaultPort mirrors the resource's hardcoded listening port.
+const postgresDefaultPort int64 = 5432
+
+var _ datasource.DataSource = &postgresServiceDataSource{}
+
+// NewPostgresServiceDataSource fetches a single Managed Postgres service by ID.
+func NewPostgresServiceDataSource() datasource.DataSource {
+	return &postgresServiceDataSource{}
```
heavycrystal, line 26: nit: I think this function does the same compiler verification as the `var _` above would do. Might apply elsewhere

pkg/resource/postgres_service.go
amogiska: I removed it altogether. The resource now always hydrates password from the server on read/create, so there's nothing to persist here; passwordCreateIntent is just {Set, Value} now.

pkg/resource/postgres_service.go
amogiska: updated them to "change (incrementing by convention)" so the code and docs agree.

pkg/resource/postgres_service.go
amogiska: USFU is the UseStateForUnknown plan modifier. Spelled it out in the comments.

pkg/datasource/postgres_service.go
amogiska: Makes sense but I'm going to leave it for consistency with the rest of the provider.

docs/resources/postgres_service.md
```diff
@@ -287,15 +295,23 @@ them. The provider reads the source at plan time and fills them in:
...
   replica. Mutually exclusive with `restore_to_point_in_time` and with
   `password` / `password_wo` (a replica inherits the primary's superuser).
   Changing or removing it **destroys and recreates** the instance as a
   standalone primary — a live replica can't be converted in place (see
   "Out-of-band changes" for the promotion exception).
+  A **live read replica cannot be modified directly**: changing `size`,
+  `ha_type`, or `tags` is a **plan-time error** ("read replica cannot be
+  modified directly"), because the server rejects any such change on a replica.
+  Resize/retag the **parent** instead, or remove `read_replica_of` first to
+  detach this into a standalone primary. (`pg_config` / `pgbouncer_config` **are**
```
heavycrystal: "detach" implies preservation of the read replica but that is not what will happen if this is done

docs/resources/postgres_service.md
```diff
@@ -100,14 +100,22 @@ description: |-
...
   password / password_wo (a replica inherits the primary's superuser).
   Changing or removing it destroys and recreates the instance as a
   standalone primary — a live replica can't be converted in place (see
-  "Out-of-band changes" for the promotion exception).restore_to_point_in_time = { source_id, restore_target } — create
+  "Out-of-band changes" for the promotion exception).
+  A live read replica cannot be modified directly: changing size,
+  ha_type, or tags is a plan-time error ("read replica cannot be
+  modified directly"), because the server rejects any such change on a replica.
+  Resize/retag the parent instead, or remove read_replica_of first to
+  detach this into a standalone primary. (pg_config / pgbouncer_config are
```
heavycrystal: same thing, "detach" implies preservation of the read replica but that is not what will happen if this is done

heavycrystal (approved):

pkg/resource/descriptions/postgres_service.md
```diff
@@ -57,13 +50,131 @@ won't lose tags when they resize or change HA mode. This is invisible
...
+
+## Passwords
+
+The superuser password can be managed three ways:
+
+- **Omit both `password` and `password_wo`** — the server generates one,
+  captured into (sensitive) state as `password`.
+- **`password`** — a value you supply; changing it rotates the password.
+  Stored in (sensitive) state.
+- **`password_wo` + `password_wo_version`** — a write-only password: the
```
nunoadrego: This is misleading. When we use `password_wo` the expectation from users is that it shouldn't be persisted in the state. What's the point of supporting it if the state contains the password (via `connection_string`)?

nunoadrego (approved):

docs/resources/postgres_service.md
amogiska: reworded

docs/resources/postgres_service.md
amogiska: reworded

PR #565, opened 2026-06-04
feat(resource): add stop attribute to clickhouse_service

MattiasMTS:
## Summary

Adds a `stop` boolean to the `clickhouse_service` resource to stop (pause) or start a service via the ClickHouse Cloud `/state` API. The name matches ClickHouse's own vocabulary (the UI **Stop/Start** button, the `stop`/`start` API command, and the `stopped` service state).

```hcl
resource "clickhouse_service" "example" {
  # ...
  stop = true   # stop (pause) the service; false starts it
}
```

## Behavior

- **Update** starts the service *before* applying other changes and stops it *after* them, so changes are applied to a running service.
- **Create** stops the service after provisioning when `stop = true`.
- **Read** maps the API state to `stop`: `stopped`/`stopping` → `true`. Note `idle` (auto-idle) is **not** treated as stopped — an idled service is still "started" and auto-resumes on query, whereas a stopped service does not.

## Warehouse semantics + plan-time validation

In a warehouse, secondary ("replica") services can be stopped independently. ClickHouse Cloud does **not** allow stopping the primary (first) service while it still has secondary services. To avoid a confusing mid-apply API error, the provider validates this at **plan time** and returns a clear, actionable diagnostic when you try to stop a primary that still has secondaries.

## API changes

- New `ChangeServiceState(serviceId, command)` client method (the `/state` endpoint).
- New `ListServices()` client method, used by the plan-time validation to detect sibling secondary services.
- `DeleteService` now uses the shared stop-command constant.
- Mock regenerated.

## Testing

- Unit tests for the service-state → `stop` mapping and for the warehouse secondary-detection helper.
- `go build`, `go vet`, `go test ./pkg/...`, `gofmt`, `golangci-lint` (v2.12.2), and `make docs-alpha` all clean.
- Locally verified end-to-end against a local provider build: stopping a warehouse primary that has a secondary is blocked at plan time, while stopping a secondary or a standalone primary works.

## Follow-up

- Add `stop` usage to the `examples/full/warehouse` example for e2e coverage (not included here).

### DIFF
--- a/docs/resources/service.md
+++ b/docs/resources/service.md
@@ -6,6 +6,8 @@ description: |-
   You can use the clickhouse_service resource to deploy ClickHouse cloud instances on supported cloud providers.
   Known limitations:
   If you create a service with warehouse_id set and then remove warehouse_id attribute completely, the provider won't detect the change. If you want to make a secondary service become primary, remove the warehouse_id and taint it before applying.If you create a service with readonly flag set to true and then remove readonly flag completely, the provider won't detect the change. If you want to make a secondary service read write, explicitly set the readonly flag to false.
+  Stopping a service:
+  Setting stop = true stops the service: its compute is fully shut down so compute billing stops (storage is still billed). Unlike auto-idle (see idle_scaling), a stopped service does NOT auto-resume on query; set stop = false to start it again.Secondary (replica) services in a warehouse can be stopped independently — stopping one only shuts down that service's compute; the primary and sibling services keep running.The primary (first) service of a warehouse cannot be stopped while any secondary services still exist: ClickHouse Cloud requires the first service to stay up. Remove the secondary services before stopping the primary.
 ---
 
 # clickhouse_service (Resource)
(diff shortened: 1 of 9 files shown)


nunoadrego:
Hey @MattiasMTS, have you considered implementing this as an [action](https://developer.hashicorp.com/terraform/language/invoke-actions)? At first glance, that seems more appropriate. Would that fit your use case?

MattiasMTS:
> Hey @MattiasMTS, have you considered implementing this as an [action](https://developer.hashicorp.com/terraform/language/invoke-actions)? At first glance, that seems more appropriate. Would that fit your use case?

Didn't know about it. I'll check! You think it makes most sense?

nunoadrego:
> > Hey @MattiasMTS, have you considered implementing this as an [action](https://developer.hashicorp.com/terraform/language/invoke-actions)? At first glance, that seems more appropriate. Would that fit your use case?
> 
> Didn't know about it. I'll check! You think it makes most sense?

I haven't used any so far, but I am aware of them. I think it's worth exploring. 🤔

I see we can [trigger it on lifecycle events](https://developer.hashicorp.com/terraform/language/invoke-actions#trigger-an-action-when-applying-a-configuration).

Or [command only](https://developer.hashicorp.com/terraform/language/invoke-actions#invoke-an-action-from-the-cli):
```
terraform plan -invoke=action.<tbd>.example
terraform apply -invoke=action.<tbd>.example
```

What do you think?

MattiasMTS:
@nunoadrego this won't be supported on opentofu tho', since tofu haven't added support for it yet. Do we think that is acceptable? We are using opentofu, so I'll probably drop this PR if that's the case 😅

PR #580, opened 2026-06-18
docs(clickpipes): note MySQL FILE_POS replication requires a feature flag

dtunikov:
## What

Creating a MySQL ClickPipe with a replication mechanism other than `GTID` (e.g. `FILE_POS`) now requires the `FT_ORG_CLICKPIPES_MYSQL_REPLICATION_SELECT` organization feature to be enabled by ClickHouse support. The Cloud API returns `BAD_REQUEST` at create time when the feature is off and the mechanism is not `GTID`.

This PR documents that constraint so users aren't surprised by the API error.

## Changes

- `pkg/resource/clickpipe.go` — extend the `replication_mechanism` schema `MarkdownDescription` (source of truth for the generated docs) to note that mechanisms other than `GTID` must be enabled for the organization.
- `docs/resources/clickpipe.md` — regenerated line for `replication_mechanism` to match the schema description.
- `examples/clickpipe/cdc_mysql/main.tf` and `README.md` — add the same caveat to the example.

## Notes

- Docs-only / description change; no behavior change in the provider. `FILE_POS` is still a valid `OneOf` value — enforcement of the feature gate lives in the Cloud API.


ilidemi (commented):
Please include MariaDB and prevalidate it on `plan`

whites11 (approved):

ilidemi (approved):

PR #590, opened 2026-07-01
Add ClickPipes object storage load options

kitop:
## Summary

Adds Terraform support for ClickPipes object storage `skip_initial_load` and `start_after`, matching the OpenAPI fields. The resource now validates that `skip_initial_load` requires `queue_url` and cannot be combined with `start_after`, maps both fields into API requests, and preserves explicit false values during state sync. Generated ClickPipe docs and focused unit coverage were updated.

## Validation

Added tests.


tpanetti (approved):
LGTM, changes look safe. I would just run an actual test against the API to make sure the settings are permeating before merging if you haven't done so already

Issue #529, opened 2026-05-06
clickhouse_clickpipe Postgres CDC: “Provider produced inconsistent result after apply: .state: was Paused, but now Snapshot” when adding a table to table_mappings


dan-jacobson-artemis:
### Description

Adding a new entry to source.postgres.table_mappings on an existing Postgres CDC clickhouse_clickpipe triggers an unrecoverable Terraform error even though the underlying API call succeeds. The pipe genuinely transitions to the Snapshot state mid-PATCH (snapshotting the new table is the API’s correct behavior), but the provider’s plugin-SDK consistency check rejects the response because state differs from what it expected.

The mutation succeeds: GET against the API after the failed apply confirms the new mapping is present and the pipe is snapshotting normally. But Terraform exits non-zero, the resource state isn’t updated, and the next plan shows the same diff and re-runs the same PATCH — looping until the snapshot completes and the pipe transitions back to Running/Paused. For large tables, this can mean many minutes of failing applies.

The provider’s Update should either treat Snapshot (and other transient states like Updating) as equivalent to the prior steady state for the inconsistency check, or omit state from the check entirely (it’s an API-driven field, not a user-driven one)

### Cloud provider

AWS

### Service tier

Not applicable

### Affected resource(s)

clickhouse_clickpipe (Postgres CDC, source.postgres.type = "rdspostgres")

### Provider version

3.14.0

### Terraform version

1.14.0

### Terraform configuration

```hcl
resource "clickhouse_clickpipe" "lakehouse_cdc" {
  name       = "lakehouse-cdc-example"
  service_id = var.clickhouse_service_id

  source = {
    postgres = {
      type     = "rdspostgres"
      host     = var.host
      port     = 5432
      database = var.database

      credentials = {
        username = "lakehouse_replication"
        password = data.aws_ssm_parameter.replication_password.value
      }

      settings = {
        replication_mode = "cdc"
        publication_name = "lakehouse_cdc"
      }

      table_mappings = [
        { source_schema_name = "public", source_table = "case_table",    target_table = "example_case_table",    table_engine = "ReplacingMergeTree" },
        { source_schema_name = "public", source_table = "finding_table", target_table = "example_finding_table", table_engine = "ReplacingMergeTree" },
        # Step 4 of the repro adds a row here.
      ]
    }
  }

  destination = {
    database = "rds_lakehouse"
  }
}
```

### Expected behavior

[...]

PR #596, opened 2026-07-06
fix(clickpipes): decide planned state after ModifyPlan repairs to avoid perpetual plan diff

### LINKED ISSUE #595: clickhouse_clickpipe: permanent "state → (known after apply)" plan diff for Postgres CDC pipes (regression of #447 since 3.16.0)
### Description

Every `terraform plan` shows our Postgres CDC ClickPipes as `will be updated in-place` with a single change: `state = "Running" -> (known after apply)`. Applying is a no-op (`Update` detects no real changes and skips the PATCH), and the next plan shows the identical diff again — it never converges.

This is the same symptom as #447, which #456 fixed with `UseStateForUnknown()`. It returned when #532 replaced that plan modifier with the custom `volatileComputedString` (shipped in v3.16.0).

Analysis from reading the provider and terraform-plugin-framework source:

- `volatileComputedString`'s refresh-only branch tests `req.State.Raw.Equal(req.Plan.Raw)`, but attribute plan modifiers run *after* the framework has applied schema defaults and run `MarkComputedNilsAsUnknown`, which has already marked `state` itself unknown whenever the proposed plan differs from prior state anywhere. Once anything fires that gate the equality can never hold, so the branch meant to suppress refresh churn is unreachable and `state` stays unknown.
- What fires the gate for us (inferred from code, not verified against raw state): our pipes predate v3.17.1, so state holds `destination.managed_table = false` while the schema default is `true`. The default injects `true` into every plan; the `ModifyPlan` carry-forward for omitted `managed_table` repairs that diff, but only after attribute modifiers ran — leaving `state -> (known after apply)` as the sole surviving change. Since `Read` preserves the stored `false` indefinitely, the diff is permanent.

### Cloud provider

AWS

### Affected resource(s)

- clickhouse_clickpipe (Postgres CDC)

### Provider version

v3.17.3 (behavior present since v3.16.0)

### Terraform version

1.15.7

### Terraform configuration

```hcl
resource "clickhouse_clickpipe" "postgres_cdc" {
  name       = "postgres-cdc"
  service_id = var.clickhouse_service_id

  source = {
    postgres = {
      host     = var.postgres_host
      port     = 5432
      database = "app"

      credentials = {
        username = "clickpipes_user"
        password = var.clickpipes_password
      }

      settings = {
        replication_mode = "cdc"
        publication_name = "clickpipes_publication"
      }

      table_mappings = [{
        source_schema_name = "public"
        source_table       = "orders"
        target_table       = "public_orders"
        table_engine       = "ReplacingMergeTree"
      }]
    }
  }

  destination = {
    database = "analytics"
  }
}
```

`destination.managed_table` is not set. The pipe was originally created under a provider version older than v3.17.1.

### Expected behavior

An unchanged pipe produces no plan diff, as in v3.15 and earlier with `UseStateForUnknown()`.

### Actual behavior

```
  # clickhouse_clickpipe.postgres_cdc will be updated in-place
  ~ resource "clickhouse_clickpipe" "postgres_cdc" {
        id             = "<redacted>"
        name           = "postgres-cdc"
      ~ state          = "Running" -> (known after apply)
        # (5 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

Applying succeeds without calling the ClickPipes API, and the same plan reappears on the next run.

### Additional context

Bisect via released source: `state` uses `UseStateForUnknown()` through v3.15.0-alpha1; from v3.16.0 onward (including v3.17.3) it uses `volatileComputedString{}`.

billykern:
Fixes #595

`volatileComputedString` distinguishes refresh churn from a real update by comparing the whole raw plan against prior state, but it runs after the framework has already marked `state` Unknown and applied schema defaults, so the comparison can never succeed once anything differed — including differences the resource's own `ModifyPlan` later repairs, such as the `managed_table` carry-forward for pipes created before v3.17.1. Those pipes plan a perpetual `state = "Running" -> (known after apply)` update that applying never clears.

The decision now happens in a `planStateAttribute` helper that runs as `ModifyPlan`'s final step, after all plan repairs: keep the prior `state` value when the plan otherwise equals prior state, so unchanged pipes collapse to no-ops; otherwise leave it Unknown so an update may still settle in a transient state (preserving the #529 consistency-check fix). The now-unused `volatileComputedString` is removed and a regression test covers both directions.

Unit tests and golangci-lint pass locally; E2E was not run (requires ClickHouse Cloud credentials).


tpanetti:
This is amazing @billykern, thanks for the contribution! I've kicked off E2E tests [here](https://github.com/ClickHouse/terraform-provider-clickhouse/actions/runs/28955555828) and will do some independent testing but overall this is looking great!

billykern:
Great, thank you @tpanetti! Please let me know anything I can help with

tpanetti (approved):
LGTM. Ran E2E and multiple manual tests and saw no state drift in planning phase!

PR #598, opened 2026-07-06
fix(postgres): retry replica create until parent has a backup

amogiska:
Creating a read replica right after its primary finishes provisioning intermittently fails with `400: Parent server is not ready for read replicas. There are no backups, yet.` — the parent can't take replicas until its first base backup completes. Measured live: the window is usually 0s but sometimes outlasts CI's apply retries, which is what made `e2e (postgres, aws)` flaky (and leaked the primaries that eventually bricked the test org — see #597).

This treats that specific 400 as retryable in `CreatePostgresReadReplica` (20s interval, ~30 min budget, context-aware), mirroring the existing `DeletePostgres` retry pattern. Any other error still fails immediately. Matching is by status + message substring since the API exposes no structured error code for this condition.

Test plan: unit tests cover retry-then-success, fail-fast on other 400s, and fail-fast on 409; `go test ./pkg/...` passes; `go build -tags alpha` clean.

### DIFF
--- a/pkg/internal/api/postgres.go
+++ b/pkg/internal/api/postgres.go
@@ -24,9 +24,11 @@ var (
 // Retry budgets. Overridable per-call via the *WithBudget / *WithInterval
 // helpers so unit tests can run in milliseconds.
 var (
-	postgresDeleteRetryInterval        = 10 * time.Second
-	postgresDeleteRetryAttempts uint64 = 90 // 90 × 10s = 15 minutes
-	postgresStatePollInterval          = 5 * time.Second
+	postgresDeleteRetryInterval            = 10 * time.Second
+	postgresDeleteRetryAttempts     uint64 = 90 // 90 × 10s = 15 minutes
+	postgresReplicaNotReadyInterval        = 20 * time.Second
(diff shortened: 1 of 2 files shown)


pkg/internal/api/postgres.go
```diff
@@ -387,24 +389,62 @@ func (c *ClientImpl) RestorePostgres(ctx context.Context, sourceId string, body
...
+// first base backup is pending: "Parent server is not ready for read
+// replicas. There are no backups, yet." Message-substring match because the
+// public API exposes no structured error code for this condition.
+func errIndicatesReplicaNotReady(err error) bool {
+	if err == nil {
+		return false
 	}
-	return &resp.Result, nil
+	msg := err.Error()
+	if !strings.HasPrefix(msg, "status: 400") {
```
serprex, line 444: can we use `errors.AsType` to have a more well typed check?

amogiska: The string pattern is the same across the repo. Happy to introduce a typed HTTPError and migrate them all in a follow-up.

serprex (approved):

PR #599, opened 2026-07-07
feat(invitation): add clickhouse_invitation resource and clickhouse_users data source

orkhanM:
## What

This PR adds two things for organization user management:

1. **`clickhouse_invitation` resource** — invite a brand-new user to a ClickHouse Cloud organization and assign their initial roles declaratively.
2. **`clickhouse_users` data source** — list all organization members without failing when a particular user is absent.

## Why

The provider could already look up existing members (`clickhouse_user`) and manage their roles (`clickhouse_role` + `clickhouse_role_assignment`), but there was **no way to onboard a user who does not exist yet**. A member only comes into existence after they accept an invitation and log in for the first time, and plain Google OAuth does not auto-associate a user with the organization — the user must be explicitly invited. Until now that invite step was UI/API-only, so the very first step of onboarding could not be expressed in Terraform.

This closes that gap and models the natural two-phase onboarding flow:

- **Pre-acceptance** → roles are granted through the invitation's `assigned_role_ids`.
- **Post-acceptance** → the user becomes a member with a `user_id`, and `clickhouse_role_assignment` takes over ongoing role management.

## `clickhouse_invitation`

Wraps the org invitations API (`POST`/`GET`/`GET {id}`/`DELETE` under `/organizations/{org}/invitations`). Design notes:

- **Immutable.** `email`, `assigned_role_ids`, and the deprecated `role` field all use `RequiresReplace` (an invitation cannot be edited server-side). `role` and `assigned_role_ids` are mutually exclusive.
- **Read reconciles against membership.** When an invitee accepts, the invitation is consumed server-side and `GET /invitations/{id}` returns 404 — indistinguishable from a revoked/expired invite. A naive "404 → remove from state" would make Terraform re-invite an already-joined member. Instead, on 404 the resource cross-checks membership by email (via `ListMembers`): if the email is now a member the invitation is treated as **fulfilled** and kept in state with no diff; otherwise it is removed so it re-creates.
- **Create idempotency.** If the email is already a member, the API 409 is treated as fulfilled rather than a hard error.
- **Config-preserving.** Only computed fields (`id`, `created_at`, `expire_at`) are read back from the server; the immutable inputs are left as configured, so the server back-populating a legacy `role` never produces a spurious replace.

```hcl
data "clickhouse_role" "member" {
  name = "Member"
}

resource "clickhouse_invitation" "alice" {
  email             = "alice@example.com"
  assigned_role_ids = [data.clickhouse_role.member.id]
}
```

## `clickhouse_users`

The singular `clickhouse_user` data source errors with "User not found" when the email is not a member — which is exactly the case for a user who has only been invited. That makes it impossible to write configuration that references a user before they accept.

`clickhouse_users` (plural, mirroring `clickhouse_roles`) lists all members and never errors on absence: a missing user is simply not in the list. This lets you reference a possibly-not-yet-existing user and guard role assignment on their existence.

```hcl
data "clickhouse_users" "all" {}

data "clickhouse_role" "member" {
  name = "Member"
}

locals {
  # null while the invite is pending; the member object once accepted
  alice = one([for u in data.clickhouse_users.all.users : u if u.email == "alice@example.com"])
}

resource "clickhouse_role_assignment" "alice_member" {
  count    = local.alice == null ? 0 : 1
  role_id  = data.clickhouse_role.member.id
  user_ids = [local.alice.id]
}
```

## Testing

- Unit tests for the invitation resource (role-set flattening / null preservation, member-by-email lookup) and `go test ./...` passing; `golangci-lint` clean; builds on both the default and `alpha` tags.
- Verified end-to-end against a real organization: applying an invitation sends the invite and populates `id`/`created_at`/`expire_at`; re-planning a pending invitation reports no changes (idempotent); `clickhouse_users` reads without error for a pending invitee and correctly reports them as not-yet-a-member.
- Docs regenerated with `tfplugindocs`.


nunoadrego:
Thank you for your contribution. We will take a look.

nunoadrego:
@orkhanM, we had a big refactor in terms of code organization recently. Can you take a look at the conflicts?

orkhanM:
@nunoadrego thanks for the heads-up — I've rebased my branch

internal/service/clickhouse/resource/invitation.go
```diff
@@ -0,0 +1,305 @@
...
+				DeprecationMessage: "Use assigned_role_ids instead.",
+				Optional:           true,
+				Validators: []validator.String{
+					stringvalidator.OneOf("admin", "developer"),
+					stringvalidator.ConflictsWith(path.MatchRoot("assigned_role_ids")),
+				},
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
```
nunoadrego, line 90:
```suggestion
			"assigned_role_ids": schema.SetAttribute{
				Description: "Set of role IDs to assign to the invited user when they accept the invitation. Look up system role IDs with the clickhouse_role data source, or use the ID of a clickhouse_role resource for custom roles. Exactly one of assigned_role_ids or role must be set, and the set must be non-empty.",
				Optional:    true,
				ElementType: types.StringType,
				Validators: []validator.Set{
					setvalidator.SizeAtLeast(1),
				},
				PlanModifiers: []planmodifier.Set{
					setplanmodifier.RequiresReplace(),
				},
			},
			"role": schema.StringAttribute{
				Description:        "Deprecated legacy organization role for the invited user (\"admin\" or \"developer\"). Use assigned_role_ids instead. Exactly one of assigned_role_ids or role must be set.",
				DeprecationMessage: "Use assigned_role_ids instead.",
				Optional:           true,
				Validators: []validator.String{
					stringvalidator.OneOf("admin", "developer"),
				},
				PlanModifiers: []planmodifier.String{
					stringplanmodifier.RequiresReplace(),
				},
			},
```

internal/service/clickhouse/resource/invitation.go
```diff
@@ -0,0 +1,305 @@
...
+				Description: "ISO-8601 timestamp of when the invitation expires.",
+				Computed:    true,
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.UseStateForUnknown(),
+				},
+			},
+		},
+	}
+}
+
```
nunoadrego, line 108: Let's add:
```go
func (r *InvitationResource) ConfigValidators(_ context.Context) []resource.ConfigValidator {
	return []resource.ConfigValidator{
		resourcevalidator.ExactlyOneOf(
			path.MatchRoot("assigned_role_ids"),
			path.MatchRoot("role"),
		),
	}
}
```

internal/service/clickhouse/datasource/users.go
```diff
@@ -0,0 +1,181 @@
...
+	client, ok := req.ProviderData.(api.Client)
+	if !ok {
+		resp.Diagnostics.AddError(
+			"Unexpected Data Source Configure Type",
+			"Expected api.Client, got something else. Please report this issue to the provider developers.",
+		)
+		return
+	}
+	d.client = client
+}
```
nunoadrego, line 46:
```suggestion
func (d *usersDataSource) Configure(_ context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
	if req.ProviderData == nil {
		return
	}
	providerData, ok := req.ProviderData.(*service.ProviderData)
	if !ok {
		resp.Diagnostics.AddError("Unexpected provider data",
			fmt.Sprintf("expected *service.ProviderData, got %T. This is a bug in the provider.", req.ProviderData))
		return
	}
	if providerData.API == nil {
		resp.Diagnostics.AddError("ClickHouse Cloud API not configured",
			"This resource requires ClickHouse Cloud credentials. Set organization_id, token_key and token_secret on the provider (or the corresponding CLICKHOUSE_* environment variables).")
		return
	}
	d.client = providerData.API
}
```

internal/service/clickhouse/resource/invitation.go
```diff
@@ -0,0 +1,305 @@
...
+	client, ok := req.ProviderData.(api.Client)
+	if !ok {
+		resp.Diagnostics.AddError(
+			"Unexpected Resource Configure Type",
+			"Expected api.Client, got something else. Please report this issue to the provider developers.",
+		)
+		return
+	}
+	r.client = client
+}
```
nunoadrego, line 122:
```suggestion
func (r *InvitationResource) Configure(_ context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
	if req.ProviderData == nil {
		return
	}
	providerData, ok := req.ProviderData.(*service.ProviderData)
	if !ok {
		resp.Diagnostics.AddError("Unexpected provider data",
			fmt.Sprintf("expected *service.ProviderData, got %T. This is a bug in the provider.", req.ProviderData))
		return
	}
	if providerData.API == nil {
		resp.Diagnostics.AddError("ClickHouse Cloud API not configured",
			"This resource requires ClickHouse Cloud credentials. Set organization_id, token_key and token_secret on the provider (or the corresponding CLICKHOUSE_* environment variables).")
		return
	}
	r.client = providerData.API
}
```

internal/service/clickhouse/resource/descriptions/invitation.md
```diff
@@ -0,0 +1,22 @@
+Use the *clickhouse_invitation* resource to invite a new user to your ClickHouse Cloud organization and assign their initial roles. A member only exists in the organization after they accept an invitation and log in for the first time, so this resource is how you onboard a brand-new user declaratively.
+
+Invitations are immutable. Changing `email`, `assigned_role_ids`, or `role` revokes the existing invitation and sends a new one (destroy + recreate).
+
+~> **Note:** This resource governs the user's roles only *at accept time*. Manage a member's roles thereafter with the `clickhouse_role_assignment` resource, using the `user_id` from the `clickhouse_user` data source. Deleting a `clickhouse_invitation` revokes a still-pending invite; it does **not** remove a user who has already accepted.
+
+When an invited user accepts, the invitation is consumed server-side. On the next refresh the provider detects that the email now belongs to a member and keeps the resource in state with no diff, so onboarding is not repeated.
+
```
nunoadrego, line 8: Also add please:

> ~> **Note:** On import, roles are read back from `assignedRoles`, so the imported state populates `assigned_role_ids` and leaves `role` null. The API still returns the deprecated `role` field, but for organizations migrated to custom roles it is frozen at its pre-migration value and does not describe the real assignment. If you import an invitation into a configuration that uses the deprecated `role` attribute, and the organization reports assigned roles, switch the configuration to `assigned_role_ids` — otherwise the next plan will revoke and re-send the invitation.

internal/service/clickhouse/resource/invitation.go
```diff
@@ -0,0 +1,305 @@
...
+		return nil, err
+	}
+	for i := range members {
+		if members[i].Email == email {
+			return &members[i], nil
+		}
+	}
+	return nil, nil
+}
+
```
nunoadrego, line 292: Also add:
```go
// invitationRolesFromAPI derives the (assigned_role_ids, role) pair for state.
// Only one may be non-null: the schema forbids both, and since both force
// replacement, setting both makes the next plan revoke the invitation.
// assignedRoles wins, because the API keeps returning the deprecated role
// frozen at its pre-migration value once an org moves to custom roles.
func invitationRolesFromAPI(invitation *api.Invitation) (types.Set, types.String, diag.Diagnostics) {
	roleIDs := make([]string, 0, len(invitation.AssignedRoles))
	for _, ar := range invitation.AssignedRoles {
		roleIDs = append(roleIDs, ar.RoleID)
	}

	roleSet, diags := invitationRoleSet(roleIDs, true)
	if diags.HasError() {
		return roleSet, types.StringNull(), diags
	}

	if !roleSet.IsNull() || invitation.Role == "" {
		return roleSet, types.StringNull(), diags
	}

	return roleSet, types.StringValue(invitation.Role), diags
}
```

internal/service/clickhouse/resource/invitation.go
```diff
@@ -0,0 +1,305 @@
...
+		for _, ar := range invitation.AssignedRoles {
+			roleIDs = append(roleIDs, ar.RoleID)
+		}
+		roleSet, d := invitationRoleSet(roleIDs, true)
+		resp.Diagnostics.Append(d...)
+		if resp.Diagnostics.HasError() {
+			return
+		}
+		state.AssignedRoleIDs = roleSet
+	}
```
nunoadrego, line 238:
```suggestion
	if state.Email.IsNull() {
		state.Email = types.StringValue(invitation.Email)

		roleSet, role, d := invitationRolesFromAPI(invitation)
		resp.Diagnostics.Append(d...)
		if resp.Diagnostics.HasError() {
			return
		}
		state.AssignedRoleIDs = roleSet
		state.Role = role
	}
```

internal/service/clickhouse/resource/invitation.go
```diff
@@ -0,0 +1,305 @@
...
+		Email:           plan.Email.ValueString(),
+		AssignedRoleIds: roleIDs,
+		Role:            plan.Role.ValueString(),
+	})
+	if err != nil {
+		// If the email already belongs to a member the API returns 409. Treat
+		// that as a fulfilled onboarding rather than a hard error: the user is
+		// already in the org, so re-inviting is unnecessary. There is no
+		// invitation to reference, so key the resource off the member's user ID;
+		// a later Read will 404 on GetInvitation and reconcile via membership.
```
nunoadrego, line 149:
```suggestion
               // The API returns 409 both when the email already belongs to a member
               // and when an invitation for it is already pending. Neither creates an
               // invitation, so no resource may enter state: there is no invitation ID
               // to store, and storing anything else (a user ID, say) yields a resource
               // that Read and Delete can never resolve. Distinguish the two cases so
               // the diagnostic names the actual remedy.
```

internal/service/clickhouse/resource/invitation.go
```diff
@@ -0,0 +1,305 @@
...
+			}
+			if member != nil {
+				tflog.Info(ctx, "Invitation target is already a member; treating as fulfilled", map[string]any{"email": plan.Email.ValueString()})
+				plan.ID = types.StringValue(member.UserID)
+				plan.CreatedAt = types.StringValue("")
+				plan.ExpireAt = types.StringValue("")
+				resp.Diagnostics.Append(resp.State.Set(ctx, plan)...)
+				return
+			}
+		}
```
nunoadrego, line 164:
```suggestion
		if api.IsConflict(err) {
			member, memberErr := r.findMemberByEmail(ctx, plan.Email.ValueString())
			if memberErr != nil {
				resp.Diagnostics.AddError("Error creating invitation", "Invitation returned a conflict and membership could not be verified: "+memberErr.Error())
				return
			}
			if member != nil {
				resp.Diagnostics.AddError("User is already an organization member",
					fmt.Sprintf("%s is already a member of this organization, so no invitation can be created. "+
						"Remove this clickhouse_invitation resource and manage the member's roles with "+
						"clickhouse_role_assignment, looking the user up with the clickhouse_users data source.",
						plan.Email.ValueString()))
				return
			}
			resp.Diagnostics.AddError("Invitation already exists",
				fmt.Sprintf("An invitation for %s is already pending in this organization. Import it with "+
					"'terraform import' instead of creating a new one.", plan.Email.ValueString()))
			return
		}
```

nunoadrego (commented):
Hey. I've reviewed the code and done some local testing. Let me know what you think.

I've encountered:
- Import drift issue.
- Allows an invite when a user is already part of the org. When you try to remove that resource, it fails because no real invitation was created.

Other than that, it looks good!

PR #602, opened 2026-07-08
feat(clickpipe): collapse provider-flavored CDC source types to base type

dtunikov:
## Context

The CDC ClickPipe UI hardcodes checks against the **base** source type (`if source.type === 'postgres'`, `=== 'mysql'`). When a pipe is created via the OpenAPI path with a provider-flavored type such as `rdspostgres`, `aurorapostgres`, `cloudsqlpostgres`, `rdsmysql`, etc., the UI can't match it — `detailsRows` comes back empty and connection settings / other panels fail to render.

This is the Terraform-provider side of the "fix-forward" approach: revert `source.<db>.type` to accept only the **base** connector type, so pipes created via TF stay renderable in the UI.

Related Slack discussion: https://clickhouse-inc.slack.com/archives/C08RWC9R5E1/p1783507863681749

## What changed

- **Validation (forcing function):** `source.postgres.type` now accepts only `postgres`; `source.mysql.type` only `mysql`/`mariadb`. Flavored values (`rdspostgres`, `rdsmysql`, …) are rejected at plan time.
- **Read normalization:** legacy flavored values returned by the API are collapsed to the base type on read (`CollapsePostgresSourceType` / `CollapseMySQLSourceType`), so existing state converges without a perpetual diff. The MySQL ↔ MariaDB engine distinction is preserved (`rdsmariadb` → `mariadb`, not `mysql`).
- **No destructive replace:** the `RequiresReplace()` plan modifier on `type` is replaced with `requiresReplaceIfBaseTypeChanges`, which only forces replacement when the **base** engine actually changes. Collapsing a flavored value to its base (`rdspostgres` → `postgres`) is an **in-place update** — no re-snapshot, no data disruption. A genuine `mysql` → `mariadb` change still requires replacement.
- Regenerated docs + unit tests for the plan modifier and collapse helpers.

## Customer migration

Existing configs with a flavored `type` will fail validation until updated. Recommended change:

```diff
 source {
   postgres {
-    type = "rdspostgres"
+    type = "postgres"
     ...
   }
 }
```

Because the collapse is treated as a non-replacing change, applying after the edit performs an in-place update — the pipe is **not** recreated. No `ignore_changes` needed.

## Open questions / follow-ups

- **Flavor visibility is dropped.** This PR removes the sub-type from `type` without a replacement field. If we want to retain provider/flavor info (per Marta's suggestion), a follow-up should add an optional `db_provider` (or similar) field once the API/OpenAPI supports it. Left out here because it depends on backend support.
- **Docs** were hand-aligned to the new schema descriptions; run `make docs` in CI to confirm they match tfplugindocs output.
- Draft pending the product decision in the thread (option 1 long-term vs. this fix-forward).

### DIFF
--- a/docs/resources/clickpipe.md
+++ b/docs/resources/clickpipe.md
@@ -401,7 +401,7 @@ Optional:
 - `port` (Number) The port of the MySQL instance. Default is 3306.
 - `skip_cert_verification` (Boolean) Skip certificate verification for the MySQL connection.
 - `tls_host` (String) TLS/SSL host for secure connections. Used to verify the server certificate.
-- `type` (String) The type of MySQL-compatible source. (`mysql`, `rdsmysql`, `auroramysql`, `planetscalevitess`, `mariadb`, `rdsmariadb`). Default is `mysql`.
+- `type` (String) The type of MySQL-compatible source. (`mysql`, `mariadb`). Default is `mysql`. Provider-flavored values such as `rdsmysql` are deprecated; use the base `mysql` or `mariadb` type.
 
 <a id="nestedatt--source--mysql--credentials"></a>
 ### Nested Schema for `source.mysql.credentials`
@@ -509,7 +509,7 @@ Optional:
 - `iam_role` (String) IAM role ARN for IAM authentication. Required when authentication is set to `iam_role`.
 - `port` (Number) The port of the Postgres instance. Default is 5432.
 - `tls_host` (String) TLS/SSL host for secure connections. Used to verify the server certificate.
-- `type` (String) The type of the Postgres source. (`postgres`, `supabase`, `neon`, `alloydb`, `planetscale`, `rdspostgres`, `aurorapostgres`, `cloudsqlpostgres`, `azurepostgres`, `crunchybridge`, `tigerdata`). Default is `postgres`.
+- `type` (String) The type of the Postgres source. (`postgres`). Default is `postgres`. Provider-flavored values such as `rdspostgres` are deprecated; use the base `postgres` type.
 
 <a id="nestedatt--source--postgres--credentials"></a>
 ### Nested Schema for `source.postgres.credentials`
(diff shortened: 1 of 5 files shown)


dtunikov:
### Update-request behavior for the collapse

Traced what happens when an existing pipe is collapsed (`rdspostgres` → `postgres`) in an in-place update:

- **type-only change** → `pipeChanged` stays false (`type` isn't in the change-detection set), so `UpdateClickPipe` is never called. The collapse is a state-only convergence; the API is never sent the new type.
- **Postgres**, even bundled with other field changes → `type` is omitted on update (`!isUpdate` guard + `type,omitempty`).
- **MySQL** → previously set `type` **unconditionally** in the source struct, so a MySQL update touching any field would PATCH the base type onto a pipe stored under a legacy flavored type. Second commit guards it with `!isUpdate` to match Postgres.

Net: neither source ever sends `type` on update, so the collapse never depends on the OpenAPI PATCH endpoint accepting a changed `type`.

docs/resources/clickpipe.md
```diff
@@ -401,7 +401,7 @@ Optional:
 - `port` (Number) The port of the MySQL instance. Default is 3306.
 - `skip_cert_verification` (Boolean) Skip certificate verification for the MySQL connection.
 - `tls_host` (String) TLS/SSL host for secure connections. Used to verify the server certificate.
-- `type` (String) The type of MySQL-compatible source. (`mysql`, `rdsmysql`, `auroramysql`, `planetscalevitess`, `mariadb`, `rdsmariadb`). Default is `mysql`.
+- `type` (String) The type of MySQL-compatible source. (`mysql`, `mariadb`). Default is `mysql`. Provider-flavored values such as `rdsmysql` are deprecated; use the base `mysql` or `mariadb` type.
```
dtunikov, line 404: it's just a draft PR to illustrate how we could deprecate flavored postgres/mysql source types without breaking existing pipes.

PR #600, opened 2026-07-07
fix(clickpipes): pause CDC pipe before editing table_mappings

david-cove:
## Problem

CDC ClickPipe sources (Postgres, MySQL, MongoDB) reject a `table_mappings`
edit unless the pipe is paused: the API returns `ClickPipe must be paused to
edit`. The provider issues the edit without pausing, so a single `terraform
apply` that adds or removes a table fails, and the user has to pause the pipe
by hand, apply, then resume. See #497.

## Fix

In `Update()`, when the plan changes `table_mappings` on a CDC source and the
pipe is currently running, pause it and wait for the paused state before
issuing the edit. The existing stop/start reconciliation then resumes the pipe
to the plan's desired state.

- `requiresPauseForEdit` is set only when `table_mappings` actually changed on a
  Postgres/MySQL/MongoDB source, so pipes are never needlessly paused for
  unrelated edits (renames, scaling changes) or for non-CDC sources.
- A deferred guard resumes the pipe if `Update()` returns an error after the
  pause, so a failed edit cannot leave the pipe stuck paused. It keys off
  whether we actually issued the pause, so a pipe that was already stopped out
  of band is left as-is.
- The Stopped/Paused check used by both the pause gate and the wait is factored
  into `isClickPipeStoppedState`, reused from the existing state-check logic.

## Testing

- New unit test `TestIsClickPipeStoppedState` covering terminal, transitional,
  and active states.
- `go build ./...`, `go vet ./...`, full `pkg/resource` unit tests pass.
- `make fmt` (golangci-lint `--fix` with gofumpt/goimports) reports 0 issues and
  no reformatting.

Fixes #497

PR #604, opened 2026-07-09
feat(ClickPipes): Pause cdc ClickPipe before table mappings edit

### LINKED ISSUE #497: Postgres Clickpipe must be paused to edit
### Description

I get an error and failed `terraform apply` when attempting to add more tables to an existing Postgres Clickpipe.

I searched the issues first and could not find any relevant open or closed issues related to a Clickpipe needing to be paused ("pause" and "paused" search terms produce `0` results).

### Cloud provider

Not applicable

### Service tier

Not applicable

### Affected resource(s)

clickhouse_clickpipe

### Provider version

3.11.1-alpha1

### Terraform version

1.11.1

### Terraform configuration

```hcl
It's pretty self-explanatory. I'm using Terragrunt v0.75.0 so I'm not sure how to produce the minimal TF config
```

### Expected behavior

The TF Provider automates pausing the Clickpipe itself before performing the update.

### Actual behavior

I get the following error from my CI/CD (Github Actions run):

```
  ╷
  │ Error: Error Updating ClickPipe
  │ 
  │   with clickhouse_clickpipe.postgres_clickpipes["pg_public_cdc"],
  │   on main.tf line 229, in resource "clickhouse_clickpipe" "postgres_clickpipes":
  │  229: resource "clickhouse_clickpipe" "postgres_clickpipes" {
  │ 
  │ Could not update ClickPipe, unexpected error: status: 400, body:
  │ {"requestId":"d3c7e480-84a0-4cd1-a877-e3397e4715af","error":"BAD_REQUEST:
  │ Postgres ClickPipe must be paused to edit","status":400}
  ╵
```

### Additional context

In order to get it to pass and do what is expected, I have to use my browser and navigate to the **Web Console > Service > Data Sources > Clickpipe > Settings** and manually pause the clickpipe.

tpanetti:
# Problem

CDC ClickPipe sources (Postgres, MySQL, MongoDB) reject a table_mappings edit unless the pipe is paused — the API returns 400: ClickPipe must be paused to edit. The provider issued the edit without pausing, so a single terraform apply that adds or removes a table failed, and users had to run two applies (stopped = true → apply → change + stopped = false → apply), doubling every CI/CD pass just to add a table.

Fixes #497
Supersedes and includes #600 — thanks @david-cove for the base implementation.

# Fix

Update() now orchestrates the pause transparently:

1. When the plan changes table_mappings on a CDC source and the pipe is not already paused, the provider issues stop and waits for a terminal paused state (Stopped/Paused) before sending the PATCH.
2. After the edit, the API auto-resumes the pipe to validate the change; the existing stop/start reconciliation then converges the pipe to the plan's declared stopped value, so no dedicated resume logic is needed on the happy path.
3. Safety net: if the PATCH is rejected with the must be paused to edit 400 even though the gate didn't fire (pipe resumed out of band, or a field the API newly requires a pause for), the provider trusts the API's verdict, pauses, and retries the edit exactly once — the same detect-and-retry shape as the idle-service wake in #603.

Pipes are never needlessly paused for unrelated edits (renames, scaling changes) or non-CDC sources.

## Failure semantics

A failed edit cannot strand a pipe in the wrong state:

- If Update() errors after the pause, a deferred guard resumes the pipe — unless the plan declares stopped = true, in which case paused already matches the user's intent and the pipe is left alone.
- If that recovery resume itself fails, the apply surfaces a ClickPipe may be left paused warning with the underlying error instead of failing silently.

## Implementation notes

- The pause+wait sequence is factored into pauseClickPipeForEdit, shared by the plan-diff gate and the 400-retry path; the Stopped/Paused check is factored into isClickPipeStoppedState, reused from the existing state-check logic.
- The stop/start reconciliation consumes the PATCH response's live state directly, which correctly handles the window where the API's auto-resume is not yet observable.

## Testing

- New Update-flow tests drive ClickPipeResource.Update end-to-end against a mocked API client and pin the exact call order:
  - happy path: stop → wait(Paused) → PATCH → start → wait(Running) → sync
  - failed edit → pipe resumed (plan declares running)
  - failed edit with stopped = true → no forced resume
  - failed recovery resume → warning surfaced
  - unanticipated must-be-paused 400 → pause + exactly one PATCH retry
- TestIsClickPipeStoppedState covers terminal, transitional, and active states.
- go build ./..., go vet ./..., gofmt, and the full test suite pass.
- Manually verified against a live service: adding a table to a running Postgres CDC pipe in a single apply.


pkg/resource/clickpipe.go
```diff
@@ -3764,12 +3775,32 @@ func convertMongoDBTableMappingModelToAPI(ctx context.Context, diagnostics *diag
 	return mapping
 }
 
+// isClickPipeStoppedState reports whether the pipe is in a terminal paused state
+// (Stopped or Paused), as opposed to a transitional Stopping/Pausing state. CDC
+// sources (Postgres/MySQL/MongoDB) only permit table_mappings edits while the pipe
+// is paused, so this gates the pause-before-edit and is the state to wait for.
+func isClickPipeStoppedState(state string) bool {
```
whites11: nitpicking, but I don't like the naming here. Can we be more semantically correct? Like what are we checking? maybe `isClickPipeEditable` or something like that?

whites11 (approved):
I still disagree with the concept, but if this is what we want let it be

pkg/resource/clickpipe.go
tpanetti: It's a little more applicable than just for editing so I'm hesitant to call it that (i.e. it's used for the wait target for user requested stops as well). I made it more explicit though `isClickPipeStoppedOrPaused`

pkg/resource/clickpipe.go
```diff
@@ -66,6 +66,17 @@ const (
...
+	// clickPipeMustBePausedError identifies the 400 the ClickPipes API returns
+	// for a pause-required edit, e.g. "BAD_REQUEST: Postgres ClickPipe must be
+	// paused to edit". Source-type prefix intentionally excluded.
+	clickPipeMustBePausedError = "must be paused to edit"
+
+	// clickPipeAlreadyRunningError identifies the 400 returned for a start
+	// command on a pipe that is already running — the API's post-edit
+	// auto-resume can race a start issued by the provider, and "already
+	// running" is exactly the state a start seeks.
+	clickPipeAlreadyRunningError = "ClickPipe is already running"
```
ilidemi, line 78: If not yet, let's at least add a comment on the sources that these message are considered a contract by the terraform provider. Hope we get support for structured errors at some point (after API versioning) so this could be more proper

ilidemi (approved):
Agreed that the proper solution would be in the product, but this one is easier to implement at the moment. The test set and the test framework look very cool

PR #607, opened 2026-07-12
feat: Add clickhouse_api_key resource (#585)

### LINKED ISSUE #585: Feature request: resource to create/manage organization API keys (clickhouse_api_key)
### Summary

Please add a resource (e.g. `clickhouse_api_key`) to create and manage ClickHouse Cloud organization API keys via Terraform.

### Use case

We provision ClickHouse Cloud services with this provider and need a dedicated, least-privilege API key for monitoring (scraping the per-service Prometheus endpoint `/v1/organizations/{org}/services/{id}/prometheus` via basic auth).

Today the provider can only reference **existing** keys: `clickhouse_role_assignment` accepts `api_key_ids` (added in #469), but there is no way to create the key itself. As a result key creation/rotation must happen out of band (Cloud console or `POST /v1/organizations/{orgId}/keys`), and the resulting `keyId`/`keySecret` have to be moved into our secret store manually. This breaks the otherwise fully-Terraform-managed lifecycle and makes least-privilege keys hard to manage at scale.

### Proposed behavior

A resource that wraps the existing Cloud API key endpoints:

- `POST /v1/organizations/{orgId}/keys` on create
- `GET/PATCH/DELETE /v1/organizations/{orgId}/keys/{keyId}` for read/update/delete

Suggested attributes: `name`, `roles`, `expire_at`, `state`, `ip_access_list` (where supported); exported `key_id` and a write-only/sensitive `key_secret` (returned only once on creation, consistent with the write-only password handling introduced in #532).

This would compose naturally with `clickhouse_role` / `clickhouse_role_assignment` to fully manage least-privilege keys in Terraform.

### References

- Cloud API: https://clickhouse.com/docs/cloud/manage/api/keys-api-reference
- Related: #469 (role management, references existing keys via `api_key_ids`)

barryib:
## What

Adds a `clickhouse_api_key` resource that creates and manages ClickHouse Cloud **organization API keys** (create/read/update/delete + import), resolving #585.

Today the provider can only *reference* existing keys via the `clickhouse_api_key_id` data source; there is no way to create one, forcing key creation/rotation out of band. This wraps the existing Cloud API key endpoints:

- Create: `POST /v1/organizations/{orgId}/keys`
- Read / Update / Delete: `GET|PATCH|DELETE /v1/organizations/{orgId}/keys/{keyId}`

It mirrors the structure of `clickhouse_role` (schema, CRUD, import, mock-based unit tests, `tfplugindocs`).

### Attributes

| Attribute | Mode | Notes |
|---|---|---|
| `name` | Required | Key name |
| `expire_at` | Optional | RFC3339; omit/clear = never expires |
| `state` | Optional + Computed | `enabled` / `disabled` |
| `ip_access_list` | Optional | list of `{ source, description }` |
| `id` / `key_id` | Computed | key ID |
| `key_suffix` | Computed | last chars of the key |
| `key_secret` | Computed, **Sensitive** | returned only at create; stored in state |

## Design decisions (feedback welcome)

**1. Secret handling — Computed + `Sensitive` (secret stored in state).**
The secret is server-generated and returned only once, at create. This PR stores it as a computed, sensitive attribute — the common pattern for generated credentials. Create sends the plain body (no `hashData`) so the server generates the secret; `GET` never returns it, so Read preserves the state value and **import cannot recover `key_secret`** (documented limitation).

I deliberately did **not** implement the write-only/ephemeral pattern from #532 (`password_wo` for ClickPipes). Note #532 is for *input* passwords, whereas this is a *server-returned* secret — a different shape that would likely need an `ephemeral resource` variant rather than a `_wo` argument. Happy to follow up with that if maintainers prefer that security posture.

**2. Roles intentionally omitted.**
The API's `roles` field is deprecated (create/update/read) in favor of `assignedRoleIds`/`assignedRoles`, and RBAC (system + custom roles) is already fully managed by `clickhouse_role_assignment` via `api_key_ids`. Adding `assignedRoleIds` here would create two ways to set the same binding. I left roles out to avoid that overlap and to not build on a deprecated field — but happy to add `assignedRoleIds` if you'd prefer the ergonomics and accept the documented overlap.

## Notes for reviewers

- **Create maps state from the `POST` response's nested `key` object** (same approach as `clickhouse_role`). This assumes the create response is fully hydrated for non-computed attributes (`ip_access_list`). A maintainer with live API access may want to confirm the `POST /keys` response echoes `ipAccessList`/`state`.
- **Clearing `expire_at`** sends an explicit `expireAt: null` on PATCH; **clearing `ip_access_list`** sends `[]`. This assumes PATCH treats those as "clear" rather than "no change" (standard PATCH semantics + nullable schema) — worth a sanity check against the live API.
- `ip_access_list` is modeled as an ordered `List` (mirrors `role`'s policy list). If the API reorders entries or normalizes CIDRs, a `Set` would avoid perpetual diffs — let me know if you'd prefer that.

## Testing

- Mock/httptest-based unit tests for the client (`CreateApiKey` asserts the request never contains `hashData`), the Read state-mapping (asserts the secret is preserved across a read), and the plan→request mapping including the clear-path (asserts `expireAt: null` / `ipAccessList: []`).
- `make fmt`, `make docs`, `make build`, and `go test ./...` all pass.

Closes #585


nunoadrego:
Thanks for the contribution. We will take a look.

pkg/resource/models/api_key_resource.go
```diff
@@ -0,0 +1,39 @@
...
+
+type ApiKeyResourceModel struct {
+	ID           types.String `tfsdk:"id"`
+	KeyID        types.String `tfsdk:"key_id"`
+	Name         types.String `tfsdk:"name"`
+	State        types.String `tfsdk:"state"`
+	ExpireAt     types.String `tfsdk:"expire_at"`
+	KeySuffix    types.String `tfsdk:"key_suffix"`
+	KeySecret    types.String `tfsdk:"key_secret"`
+	IpAccessList types.List   `tfsdk:"ip_access_list"`
```
nunoadrego: Follow the existing name as in the service resource.

```suggestion
	IpAccessList types.List   `tfsdk:"ip_access"`
```

pkg/resource/api_key.go
```diff
@@ -0,0 +1,365 @@
...
+			"name": schema.StringAttribute{
+				Description: "Name of the API key.",
+				Required:    true,
+			},
+			"state": schema.StringAttribute{
+				Description: "State of the key: 'enabled' or 'disabled'.",
+				Optional:    true,
+				Computed:    true,
+				Validators: []validator.String{
+					stringvalidator.OneOf("enabled", "disabled"),
```
nunoadrego: Let's do something similar to: https://github.com/ClickHouse/terraform-provider-clickhouse/blob/2c8c3b77671345092cb66a5fdfed5bfb184dae24/pkg/resource/role.go#L102

nunoadrego (commented):
Some feedback inline. Can we also reuse `models.IPAccessList`?

pkg/resource/models/api_key_resource.go
barryib: Done

pkg/resource/api_key.go
barryib: Done as suggested.

barryib:
Thanks for the feedbacks @nunoadrego. I addressed your comments.

pkg/resource/models/api_key_resource.go
barryib: Also dropped the duplicate entry model and reused `models.IPAccessList` (answering the review comment). The attribute is now `ip_access`, matching the service resource. Note: the branch was re-homed onto the #613 restructure, so this now lives at `internal/service/clickhouse/resource/models/api_key_resource.go` and the diff collapsed to a single commit — the old line anchors moved.

pkg/resource/api_key.go
barryib: Extracted the values into `api.ApiKeyStateValues` and the validator now uses `stringvalidator.OneOf(api.ApiKeyStateValues...)`, mirroring `role.go`. After the #613 re-home this lives at `internal/service/clickhouse/resource/api_key.go` (branch collapsed to a single commit, so the original line anchor moved).

barryib:
Documented import support: added `examples/resources/clickhouse_api_key/import.sh` (so the docs now render an **Import** section), and a unit test covering the import path — after `terraform import` the key is imported by ID and all fields are hydrated from the API, except `key_secret`, which stays empty since the API only returns it at creation.

internal/service/clickhouse/resource/api_key.go
```diff
@@ -0,0 +1,365 @@
...
+	if !ok {
+		resp.Diagnostics.AddError(
+			"Unexpected Resource Configure Type",
+			"Expected api.Client, got something else. Please report this issue to the provider developers.",
+		)
+		return
+	}
+
+	r.client = client
+}
```
nunoadrego, line 134: Related to recent changes in the provider

```suggestion
func (r *ApiKeyResource) Configure(_ context.Context, req resource.ConfigureRequest, resp *resource.ConfigureResponse) {
	if req.ProviderData == nil {
		return
	}

	providerData, ok := req.ProviderData.(*service.ProviderData)
	if !ok {
		resp.Diagnostics.AddError("Unexpected provider data",
			fmt.Sprintf("expected *service.ProviderData, got %T. This is a bug in the provider.", req.ProviderData))
		return
	}
	if providerData.API == nil {
		resp.Diagnostics.AddError("ClickHouse Cloud API not configured",
			"This resource requires ClickHouse Cloud credentials. Set organization_id, token_key and token_secret on the provider (or the corresponding CLICKHOUSE_* environment variables).")
		return
	}
	r.client = providerData.API
}
```

nunoadrego (commented):
Regarding this:

> 2. Roles intentionally omitted.
The API's roles field is deprecated (create/update/read) in favor of assignedRoleIds/assignedRoles, and RBAC (system + custom roles) is already fully managed by clickhouse_role_assignment via api_key_ids. Adding assignedRoleIds here would create two ways to set the same binding. I left roles out to avoid that overlap and to not build on a deprecated field — but happy to add assignedRoleIds if you'd prefer the ergonomics and accept the documented overlap.

Currently our api for a V2 org returns: `Could not create API key: status: 400, body: {"requestId":"","error":"BAD_REQUEST: 'assignedRoleIds' is required for V2 organizations.","status":400}`

So ideally we should support both `roles` and `assignedRoleIds`, or just V2, and add a note in the resource that it's only compatible with V2.

This is the behavior:
```
 * - Both provided → silently use the field matching the org's mode, ignore the other
 * - Neither provided → BAD_REQUEST
 * - V2 org + only roles → BAD_REQUEST ("Use assignedRoleIds instead")
 * - V1 org + only assignedRoleIds → BAD_REQUEST ("Use roles instead")
```

barryib:
@nunoadrego Thanks for your tests and reviews.

I tested this empirically against a real org rather than guessing from the OpenAPI spec, since the spec doesn't encode this validation at all:

```
neither field sent → 400: "'assignedRoleIds' is required for V2 organizations."
assignedRoleIds: []  → 200 OK, key created
roles: [] → 400: "roles array length < minLength. Array length: 0, minLength: 1"
roles: [] + assignedRoleIds: [] together → 400, same minLength error on roles
```

I don't actually know whether this test org is V1 or V2 — I have no way to check that from the API or docs (no field on the org resource, no separate API version, nothing in the public docs). The "required for V2 organizations" wording matches what you reported, so I'm assuming it's the same code path, but I can't confirm the org's actual classification on my end. If you can check by org ID, let me know and I'll share it.

Regardless of that org's actual type, here's what the responses show: `roles` enforces `minLength: 1` unconditionally whenever it's present in the request — so the "both provided → silently use whichever matches the org's mode" behavior doesn't apply when the other field is empty; sending `roles: [] `fails regardless of which field the org actually needs. That rules out sending both fields empty as a fix.

The one thing that does work cleanly on this org: always send `assignedRoleIds: []`, and never send roles at all. `clickhouse_role_assignment` stays the sole owner of real role bindings, no schema changes, no overlap. Given roles is already flagged deprecated in the openapi [spec](https://clickhouse.com/docs/products/cloud/api-reference/openapi/create-key), I'd guess V1 is being sunset anyway, which makes the reasonable bet either way.

The gap I can't close from here: if this org turns out to be V1, and `minLength: 1` on `roles` really is unconditional, that would mean V1 orgs have no way to create a key with zero role bindings — this resource would 400 immediately on any V1 org. Given roles is already marked deprecated on your side, I'd guess V1 is being phased out, but that's your call: is excluding V1 organizations from this resource acceptable for merge? If yes, I'll ship the `assignedRoleIds: []`-only fix and document the V2 requirement. If no, I'd need a V1 org to test against, or we fall back to exposing `roles/assigned_role_ids` as a real attribute on the resource.

nunoadrego:
Hello @barryib and anyone following this PR. I've checked internally, and we would prefer to make this resource available once all orgs are in V2. It shouldn't take longer than the end of this month.

barryib:
Hi @nunoadrego, checking in on this one. You mentioned the org-wide migration to V2 shouldn't take longer than end of August; just following up now that we're a bit past that. Any update on timing?

PR #609, opened 2026-07-12
feat: add clickhouse_service and clickhouse_services data sources (#608)

### LINKED ISSUE #608: Add clickhouse_service and clickhouse_services data sources
### Feature request

The provider has a `clickhouse_service` **resource** but no matching **data sources** to look up existing services. This makes it hard to reference a service managed elsewhere (another workspace/state, or created outside Terraform) without importing it.

### Proposal
Add two read-only data sources, mirroring the existing `clickhouse_postgres_service` / `clickhouse_postgres_services` pattern:

- **`clickhouse_service`** — look up a single service by `id`. Exposes the full read surface the API returns (identity, region/tier/state, scaling, endpoints, IP access, private endpoint config, encryption, tags, `clickhouse_version`, `created_at`), excluding secrets.
- **`clickhouse_services`** — list all services in the organization, with optional tag-based filtering (`filter=tag:Key=Value`, the API's native filter). Each item carries the same full field set.

Backed by the existing `GET /organizations/{orgId}/services/{id}` and `GET /organizations/{orgId}/services` endpoints. Mock-based unit tests, generated docs + examples.

I'm happy to contribute this (PR to follow).

barryib:
## Summary

Adds two read-only data sources for ClickHouse Cloud services, mirroring the existing `clickhouse_postgres_service` / `clickhouse_postgres_services` pattern:

- **`clickhouse_service`** — look up a single service by `id`.
- **`clickhouse_services`** — list all services in the organization, with optional tag filtering.

Closes #608.

## What's included

**API client** (`pkg/internal/api`)
- New `ListServices(ctx, filters []string) ([]Service, error)` backed by `GET /organizations/{orgId}/services`, sending the API's native tag filter as repeated `filter=tag:Key=Value` query params.
- Two new `Service` fields exposed by the API but previously unmodeled: `clickhouseVersion`, `createdAt`.

**Data sources** (`pkg/datasource`)
- Both share one object type and one `api.Service` → object mapping, so the field surface stays consistent between them.
- `clickhouse_service`: `id` (required) + full read surface (identity, region/tier/state, scaling, endpoints, IP access, encryption, tags, `clickhouse_version`, `created_at`).
- `clickhouse_services`: optional `tags` map input (→ `filter=tag:K=V`); returns every service as a list of the same object. Empty result yields a known empty list, not null.

**Tests** — mock/httptest unit tests (no `TF_ACC`, matching the repo):
- `ListServices` happy path, tag-filter encoding, and API error (httptest + go-cmp).
- Mapping happy path incl. nested endpoints/ip_access and pointer/null handling.
- A schema ↔ object-type consistency test that fails on any drift between the schema attributes and the state object (the class of mismatch that otherwise only surfaces at plan time).
- `obj.As` round-trip and the plural list-mapping (length + known-empty).

## Design notes for reviewers

- **Field surface is intentionally a curated read subset.** Excluded: passwords / password hashes (secrets, never returned by GET); `query_api_endpoints` and `backup_configuration` (each requires a separate sub-resource call — out of scope here).
- **`private_endpoint_config` is intentionally omitted from both.** It isn't part of the base service payload — `GetService` enriches it via a separate `/privateEndpointConfig` call, which the list endpoint can't replicate without an N+1. Rather than expose a field that would be populated for the single-service data source but always null for the list, it's dropped from both for parity. `private_endpoint_ids` (part of the base payload) is retained on both.
- **`GetService` reuse (minor):** the singular data source reuses the existing `GetService`, which also fetches backup configuration (primary services) and query endpoints. Those results are not exposed here, but the calls still happen, so a single-service read depends on those sub-resources being reachable. Happy to special-case this if maintainers prefer a leaner read.
- The generated `client_mock.go` shows a few blank-line-only removals in unrelated methods — that's the repo's pre-commit `golangci-lint --fix` whitespace behavior, not a hand edit.


nunoadrego:
Thanks for the contribution. We will take a look.

internal/service/clickhouse/datasource/service.go
```diff
@@ -0,0 +1,338 @@
...
+		"endpoints":                          endpoints,
+		"tags":                               tags,
+	})
+	diags.Append(d...)
+	return obj, diags
+}
+
+// serviceComputedAttributes returns the computed attribute schema shared by the
+// singular data source (top level) and the plural data source (nested).
+func serviceComputedAttributes() map[string]schema.Attribute {
```
nunoadrego, line 216: Can we add descriptions similar to what we do [here](https://github.com/ClickHouse/terraform-provider-clickhouse/blob/2c8c3b77671345092cb66a5fdfed5bfb184dae24/pkg/datasource/postgres_service.go#L75)?

internal/service/clickhouse/datasource/service.go
```diff
@@ -0,0 +1,338 @@
...
+		"state":                              types.StringType,
+		"clickhouse_version":                 types.StringType,
+		"created_at":                         types.StringType,
+		"release_channel":                    types.StringType,
+		"is_primary":                         types.BoolType,
+		"readonly":                           types.BoolType,
+		"compliance_type":                    types.StringType,
+		"byoc_id":                            types.StringType,
+		"warehouse_id":                       types.StringType,
+		"num_replicas":                       types.Int64Type,
```
nunoadrego, line 58: Latest version in main now includes: MinReplicas, MaxReplicas, and AutoscalingMode. Can we add them?

internal/service/clickhouse/datasource/service.go
```diff
@@ -0,0 +1,338 @@
...
+		"iam_role":                           types.StringType,
+		"ip_access":                          types.ListType{ElemType: ipAccessObjectType()},
+		"private_endpoint_ids":               types.ListType{ElemType: types.StringType},
+		"encryption_key":                     types.StringType,
+		"encryption_assumed_role_identifier": types.StringType,
+		"has_transparent_data_encryption":    types.BoolType,
+		"transparent_data_encryption_key_id": types.StringType,
+		"encryption_role_id":                 types.StringType,
+		"enable_core_dumps":                  types.BoolType,
+		"endpoints":                          types.ListType{ElemType: endpointObjectType()},
```
nunoadrego, line 77: The resource models endpoints as a SingleNestedAttribute keyed by protocol (nativesecure/https/mysql) — resource/service.go:244-. The data source models them as a list of {protocol, host, port}.

nunoadrego (commented):
Left some comments inline. Also:

> GetService reuse (minor): the singular data source reuses the existing GetService, which also fetches backup configuration (primary services) and query endpoints. Those results are not exposed here, but the calls still happen, so a single-service read depends on those sub-resources being reachable. Happy to special-case this if maintainers prefer a leaner read.

I would prefer to use only the API endpoint needed with a new method like `func (c *ClientImpl) GetServiceBase `.

internal/service/clickhouse/datasource/service.go
barryib: Done in 3cb3665 — added `Description` to every attribute (top-level and nested), mirroring `postgres_service.go`.

internal/service/clickhouse/datasource/service.go
barryib: Done in 3cb3665 — added `autoscaling_mode`, `min_replicas`, and `max_replicas`.

internal/service/clickhouse/datasource/service.go
barryib: Kept as a flat list of `{protocol, host, port}` intentionally: it mirrors the API's `endpoints` array and stays lossless for any protocol the API returns. Matching the resource's protocol-keyed `nativesecure`/`https`/`mysql` object would silently drop any endpoint outside those three protocols. Happy to switch to the nested shape if you'd prefer provider consistency and are OK with that limitation — let me know.

barryib:
Thanks for the review! Addressed in 3cb3665:

- **Lean API call:** added `GetServiceBase` (raw `GET /services/{id}` only). `GetService` now reuses it and layers on the enrichment calls, and the `clickhouse_service` data source uses `GetServiceBase` — so it no longer makes the unused private-endpoint-config / backup / query-endpoint requests. Added a test asserting `GetServiceBase` makes exactly one HTTP call.
- **Autoscaling fields:** added `autoscaling_mode`, `min_replicas`, `max_replicas`.
- **Descriptions:** added to every attribute (docs regenerated).
- **Endpoints:** replied inline — kept the flat list to stay lossless; the nested protocol-keyed shape would drop endpoints outside nativesecure/https/mysql. Happy to change if you prefer.

barryib:
Rebased onto `main` to pick up the internal/service group restructure (#613). The data sources now live in `internal/service/clickhouse/datasource/` and register via the clickhouse `ServicePackage`; the API additions are under `internal/api`. Added local `strOrNull`/`apiTagsToStringMap` helpers to the clickhouse datasource package (the postgres datasource that previously provided them is now a separate service group) and bumped the registry golden data-source count. Build + full test suite green.

internal/service/clickhouse/datasource/services.go
```diff
@@ -0,0 +1,126 @@
...
+	if req.ProviderData == nil {
+		return
+	}
+	client, ok := req.ProviderData.(api.Client)
+	if !ok {
+		resp.Diagnostics.AddError("Unexpected Data Source Configure Type", "Expected api.Client, got something else. Please report this issue to the provider developers.")
+		return
+	}
+	d.client = client
+}
```
nunoadrego, line 81:
```suggestion
func (d *servicesDataSource) Configure(_ context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
	if req.ProviderData == nil {
		return
	}
	providerData, ok := req.ProviderData.(*service.ProviderData)
	if !ok {
		resp.Diagnostics.AddError("Unexpected provider data",
			fmt.Sprintf("expected *service.ProviderData, got %T. This is a bug in the provider.", req.ProviderData))
		return
	}
	if providerData.API == nil {
		resp.Diagnostics.AddError("ClickHouse Cloud API not configured",
			"This resource requires ClickHouse Cloud credentials. Set organization_id, token_key and token_secret on the provider (or the corresponding CLICKHOUSE_* environment variables).")
		return
	}
	d.client = providerData.API
}
```

internal/service/clickhouse/datasource/service.go
```diff
@@ -0,0 +1,372 @@
...
+	if req.ProviderData == nil {
+		return
+	}
+	client, ok := req.ProviderData.(api.Client)
+	if !ok {
+		resp.Diagnostics.AddError("Unexpected Data Source Configure Type", "Expected api.Client, got something else. Please report this issue to the provider developers.")
+		return
+	}
+	d.client = client
+}
```
nunoadrego, line 297:
```suggestion
func (d *serviceDataSource) Configure(_ context.Context, req datasource.ConfigureRequest, resp *datasource.ConfigureResponse) {
	if req.ProviderData == nil {
		return
	}
	providerData, ok := req.ProviderData.(*service.ProviderData)
	if !ok {
		resp.Diagnostics.AddError("Unexpected provider data",
			fmt.Sprintf("expected *service.ProviderData, got %T. This is a bug in the provider.", req.ProviderData))
		return
	}
	if providerData.API == nil {
		resp.Diagnostics.AddError("ClickHouse Cloud API not configured",
			"This resource requires ClickHouse Cloud credentials. Set organization_id, token_key and token_secret on the provider (or the corresponding CLICKHOUSE_* environment variables).")
		return
	}
	d.client = providerData.API
}
```

nunoadrego (commented):
This is due to the recent refactor. Also, `make test` is failing.

barryib:
> This is due to the recent refactor. Also, `make test` is failing.

I apply your suggestions and `make test` is green now.

nunoadrego (approved):

PR #612, opened 2026-07-14
fix(scheduled_scaling): make entries applyable with idle_scaling = false

### LINKED ISSUE #611: clickhouse_service_scheduled_scaling: "inconsistent result after apply" for entries with idle_scaling = false
### Description

# clickhouse_service_scheduled_scaling: "inconsistent result after apply" for entries with idle_scaling = false

## Description

`clickhouse_service_scheduled_scaling` cannot be applied for any entry with `idle_scaling = false`. `terraform apply` fails with *"Provider produced inconsistent result after apply … does not correlate with any element in actual"*. Since disabled idle scaling is the common case for fixed-size scheduled windows, the resource is effectively unusable for those entries.

## Cloud provider

AWS

## Service tier

Not applicable

## Affected resource(s)

`clickhouse_service_scheduled_scaling`

## Provider version

`3.18.0` (resource introduced in 3.17.0; affects all versions that ship it)

## Terraform version

`1.12.2`

## Terraform configuration

```hcl
resource "clickhouse_service_scheduled_scaling" "reader" {
  service_id = clickhouse_service.reader.id

  entries = [
    {
      name                  = "Business Hours"
      weekdays              = [0, 1, 2, 3, 4, 5, 6]
      start_hour_utc        = 5
      end_hour_utc          = 19
      min_replicas          = 10
      max_replicas          = 10
      min_replica_memory_gb = 236
      max_replica_memory_gb = 236
      idle_scaling          = false
    },
  ]
}
```

## Expected behavior

The schedule is created and the apply completes cleanly, with the entry stored as written (`idle_scaling = false`).

## Actual behavior

The API call succeeds, but the apply then aborts with a provider-consistency error:

```
Error: Provider produced inconsistent result after apply

When applying changes to clickhouse_service_scheduled_scaling.reader, provider
"provider[\"registry.terraform.io/clickhouse/clickhouse\"]" produced an unexpected
new value: .entries: planned set element cty.ObjectVal(map[string]cty.Value{
"end_hour_utc":cty.NumberIntVal(19), "idle_scaling":cty.False,
"idle_timeout_minutes":cty.NumberIntVal(5), "max_replica_memory_gb":cty.NumberIntVal(236),
"max_replicas":cty.NumberIntVal(10), "min_replica_memory_gb":cty.NumberIntVal(236),
"min_replicas":cty.NumberIntVal(10), "name":cty.StringVal("Business Hours"),
"start_hour_utc":cty.NumberIntVal(5), "weekdays":cty.SetVal([]cty.Value{...})})
does not correlate with any element in actual.

This is a bug in the provider, which should be reported in the provider's own
issue tracker.
```

The same happens whether `idle_timeout_minutes` is set or omitted. Retrying loops: a subsequent plan shows `-/+ replacement` because the failed apply leaves the resource tainted in state.

## Additional context

**Root cause (from reading the source at `v3.18.0`):**

`entries` is a `SetNestedAttribute` whose nested fields are `Optional + Computed`. Two problems compound:

1. Terraform correlates **set** elements by a hash of their contents. Any nested field left unset is `(known after apply)` at plan time; a set element containing an unknown can't be hash-matched to a returned element → "does not correlate."
2. For a non-idle entry the server does not echo the idle fields, and `apiEntryToModel` maps them back with `nil → null` (`boolPtrToValue` / `int64PtrToValue` in `pkg/resource/service_scheduled_scaling.go`). So the applied element differs from the planned one (`idle_scaling: false` → `null`) even when every field is specified.

The existing round-trip test (`TestRoundTrip_NoServerNormalization`) doesn't catch this because it only exercises `idle_scaling = true`, where the server preserves the idle fields.

**Likely fix:** model `entries` as a `ListNestedAttribute` (lists correlate by index and tolerate Computed fields resolving after apply) and map a nil `idle_scaling` to an explicit `false`.

### Cloud provider

AWS

### Service tier

Production

### Affected resource(s)

_No response_

### Provider version

3.18.0

### Terraform version

1.12.2

### Terraform configuration

```hcl
resource "clickhouse_service_scheduled_scaling" "reader" {
  service_id = clickhouse_service.reader.id

  entries = [
    {
      name                  = "Business Hours"
      weekdays              = [0, 1, 2, 3, 4, 5, 6]
      start_hour_utc        = 5
      end_hour_utc          = 19
      min_replicas          = 10
      max_replicas          = 10
      min_replica_memory_gb = 236
      max_replica_memory_gb = 236
      idle_scaling          = false
    },
  ]
}
```

### Expected behavior

Created correctly.

### Actual behavior

Error: Provider produced inconsistent result after apply
When applying changes to clickhouse_service_scheduled_scaling.reader, provider
"provider[\"registry.terraform.io/clickhouse/clickhouse\"]" produced an unexpected
new value: .entries: planned set element cty.ObjectVal(map[string]cty.Value{
"end_hour_utc":cty.NumberIntVal(19), "idle_scaling":cty.False,
"idle_timeout_minutes":cty.NumberIntVal(5), "max_replica_memory_gb":cty.NumberIntVal(236),
"max_replicas":cty.NumberIntVal(10), "min_replica_memory_gb":cty.NumberIntVal(236),
"min_replicas":cty.NumberIntVal(10), "name":cty.StringVal("Business Hours"),
"start_hour_utc":cty.NumberIntVal(5), "weekdays":cty.SetVal([]cty.Value{...})})
does not correlate with any element in actual.
This is a bug in the provider, which should be reported in the provider's own
issue tracker.

### Additional context

_No response_

serdec:
Fixes #611.

## Problem

`clickhouse_service_scheduled_scaling` could not be applied for entries with a fixed replica count (`min_replicas == max_replicas`, the only shape the resource supports). `terraform apply` aborted with:

> Provider produced inconsistent result after apply … `.entries`: planned set element … does not correlate with any element in actual.

## Root cause

The primary cause is **API-side** (UC-1252, fixed in ClickHouse/control-plane#35956): after UC-1173, the schedule API stopped echoing `minReplicas`/`maxReplicas` for vertical entries, returning the fixed count as `numReplicas` — a field the provider does not model. The provider therefore read the replica bounds back as `null`, which cannot correlate with the planned `min/max` values.

The provider-side amplifier is that `entries` was a `SetNestedAttribute`: Terraform correlates set elements by value, so *any* server-side normalization of *any* field in an element (or an `Optional + Computed` field left unset and planned as unknown) risks a fatal "does not correlate" instead of a benign in-place update.

An earlier revision of this PR attributed the mismatch to the server dropping `idleScaling`/`idleTimeoutMinutes` for non-idle entries. Verified against the live API: it does not — those fields are echoed; the replica band was the culprit.

## What this PR changes

With the API fix in control-plane#35956, the released provider round-trips correctly again with no changes (verified). This PR is hardening so the resource cannot fail this way again for other/future response normalizations:

- **Model `entries` as a `ListNestedAttribute`.** Lists correlate by index, so a normalized or late-resolving field produces an in-place diff at worst, never an "inconsistent result after apply" error.
- **Reconcile the server response against the plan in Create/Update** (`applyScheduleToStateWithPlan` / `reconcileEntriesWithPlan`): values the user set explicitly are kept from the plan; fields left unset are resolved from the server response. This holds regardless of which fields the server echoes.
- **On Read, map a missing `idle_scaling` to its effective value `false`** so a refresh never reports drift on it.
- **Add a v0 → v1 state upgrader** (set → list layout) so existing state migrates cleanly; schema version bumped to 1.
- Update the resource description and generated docs (set → list).

## Testing

Unit tests cover the reconcile logic in both directions (explicit plan values survive normalization; unset fields resolve from the server), the Read-path `idle_scaling` default, and the v0 → v1 upgrader.

Verified end-to-end with real `terraform` runs:

- **Against the live staging API (current, pre-#35956 behavior):** the #611 config fails on `main` with the exact reported error and applies cleanly with this PR; v0 state written by `main` upgrades in place; import works. (Note: against the pre-#35956 API, a refresh still nulls `min_replicas`/`max_replicas` — resolved by the API-side fix rather than provider-side mapping.)
- **Against a mock implementing the post-#35956 response shape:** both `main` and this PR apply cleanly and plans converge, including `idle_scaling = false` with an explicit `idle_timeout_minutes`, and the v0 → v1 upgrade path.

`go build ./...`, `go test ./...`, `go vet ./...`, and `golangci-lint run` all pass.

## Rollout note

This PR is independent of control-plane#35956 in either order, but the provider release fully converges (no refresh diff on `min_replicas`/`max_replicas`) only once #35956 is deployed to prod.

---
_Generated by [Claude Code](https://claude.ai/code/session_012fxjy91WrDMdDCnDmSBSXf)_

### DIFF
--- a/pkg/resource/service_scheduled_scaling.go
+++ b/pkg/resource/service_scheduled_scaling.go
@@ -284,7 +284,7 @@ func (r *ServiceScheduledScalingResource) Read(ctx context.Context, req resource
 	}
 
 	state.ID = state.ServiceID
-	resp.Diagnostics.Append(applyScheduleToState(schedule, &state)...)
+	resp.Diagnostics.Append(applyScheduleToState(ctx, schedule, &state)...)
 	if resp.Diagnostics.HasError() {
 		return
 	}
@@ -412,7 +412,10 @@ func (r *ServiceScheduledScalingResource) UpgradeState(_ context.Context) map[in
 				entries := types.ListNull(entryType)
 				if !prior.Entries.IsNull() && !prior.Entries.IsUnknown() {
 					// The set's elements are already values of the list's
-					// element type; re-wrap them directly.
+					// element type; re-wrap them directly. This relies on the
+					// v0 and v1 entry object types being identical — if a
+					// future schema version changes the entry type, this
+					// upgrader must decode and convert instead.
 					list, d := types.ListValue(entryType, prior.Entries.Elements())
 					resp.Diagnostics.Append(d...)
 					if resp.Diagnostics.HasError() {
@@ -569,17 +572,56 @@ func planEntriesToAPI(ctx context.Context, entriesList types.List) ([]api.AutoSc
 // applyScheduleToState maps an API AutoScalingSchedule response into the
 // Terraform state model, taking entry contents from the server response. Used
 // on Read, where no plan is available to reconcile against.
-func applyScheduleToState(schedule *api.AutoScalingSchedule, state *models.ServiceScheduledScalingResourceModel) diag.Diagnostics {
+//
+// Because entries is a list, element identity is positional: writing the
+// response order verbatim would surface a spurious, non-converging reorder
+// diff after any refresh where the server returns entries in a different
+// order than the config. Entries are therefore reordered to match the prior
+// state (correlated by their identity key); entries not in prior state are
+// appended in server order. On import, prior state is empty and server order
+// is kept.
+func applyScheduleToState(ctx context.Context, schedule *api.AutoScalingSchedule, state *models.ServiceScheduledScalingResourceModel) diag.Diagnostics {
 	var diags diag.Diagnostics
 
-	entryValues := make([]attr.Value, len(schedule.Entries))
+	serverModels := make([]models.ScheduledScalingEntryModel, len(schedule.Entries))
 	for i, e := range schedule.Entries {
 		entryModel, d := apiEntryToModel(e)
 		diags.Append(d...)
 		if diags.HasError() {
 			return diags
 		}
-		entryValues[i] = entryModel.ObjectValue()
+		serverModels[i] = entryModel
+	}
+
+	var priorModels []models.ScheduledScalingEntryModel
+	if !state.Entries.IsNull() && !state.Entries.IsUnknown() {
+		diags.Append(state.Entries.ElementsAs(ctx, &priorModels, false)...)
+		if diags.HasError() {
+			return diags
+		}
+	}
+
+	ordered := make([]models.ScheduledScalingEntryModel, 0, len(serverModels))
+	used := make([]bool, len(serverModels))
+	for _, pm := range priorModels {
+		key := modelEntryKey(pm)
+		for j, sm := range serverModels {
+			if !used[j] && modelEntryKey(sm) == key {
+				ordered = append(ordered, sm)
+				used[j] = true
+				break
+			}
+		}
+	}
+	for j, sm := range serverModels {
+		if !used[j] {
+			ordered = append(ordered, sm)
+		}
+	}
+
+	entryValues := make([]attr.Value, len(ordered))
+	for i, em := range ordered {
+		entryValues[i] = em.ObjectValue()
 	}
 	entriesList, d := types.ListValue(models.ScheduledScalingEntryModel{}.ObjectType(), entryValues)
 	diags.Append(d...)
@@ -590,6 +632,27 @@ func applyScheduleToState(schedule *api.AutoScalingSchedule, state *models.Servi
 
 	applyBaseConfigToState(schedule, state)
 	return diags
+}
+
+// apiEntryKey / modelEntryKey compute a schedule entry's identity from its
+// user-declared fields: name plus the window itself (sorted weekdays and
+// hours). Names alone are not guaranteed unique, but two entries cannot cover
+// the same window (the server rejects overlaps), so the combined key is.
+func apiEntryKey(e api.AutoScalingScheduleEntry) string {
+	wd := append([]int(nil), e.Weekdays...)
+	sort.Ints(wd)
+	return fmt.Sprintf("%s|%v|%d|%d", e.Name, wd, e.StartHourUtc, e.EndHourUtc)
+}
+
+func modelEntryKey(m models.ScheduledScalingEntryModel) string {
+	wd := make([]int, 0, len(m.Weekdays.Elements()))
+	for _, v := range m.Weekdays.Elements() {
+		if iv, ok := v.(types.Int64); ok && !iv.IsNull() && !iv.IsUnknown() {
+			wd = append(wd, int(iv.ValueInt64()))
+		}
+	}
+	sort.Ints(wd)
+	return fmt.Sprintf("%s|%v|%d|%d", m.Name.ValueString(), wd, m.StartHourUtc.ValueInt64(), m.EndHourUtc.ValueInt64())
 }
 
 // applyScheduleToStateWithPlan is the Create/Update counterpart of
@@ -607,10 +670,10 @@ func applyScheduleToState(schedule *api.AutoScalingSchedule, state *models.Servi
 // Each planned value the user set (known in the plan) is kept; only fields the
 // user left unset (unknown) resolve from the server's echoed value. Entries
 // correlate to the server response by index — the POST replaces the full
-// schedule and the server preserves entry order — but the name is verified
-// before merging so a reordered response can never fill unset fields from the
-// wrong entry (they resolve to null instead, which is a valid outcome for an
-// unknown).
+// schedule and the server preserves entry order — but the entry's identity key
+// (name + window, see apiEntryKey) is verified before merging so a reordered
+// response can never fill unset fields from the wrong entry (they resolve to
+// null instead, which is a valid outcome for an unknown).
 func applyScheduleToStateWithPlan(schedule *api.AutoScalingSchedule, planModels []models.ScheduledScalingEntryModel, state *models.ServiceScheduledScalingResourceModel) diag.Diagnostics {
 	var diags diag.Diagnostics
 
@@ -619,7 +682,7 @@ func applyScheduleToStateWithPlan(schedule *api.AutoScalingSchedule, planModels 
 		// The zero model is all-null: the fallback when the server returned no
 		// matching entry at this index.
 		var server models.ScheduledScalingEntryModel
-		if i < len(schedule.Entries) && schedule.Entries[i].Name == pm.Name.ValueString() {
+		if i < len(schedule.Entries) && apiEntryKey(schedule.Entries[i]) == modelEntryKey(pm) {
 			sm, d := apiEntryToModel(schedule.Entries[i])
 			diags.Append(d...)
 			if diags.HasError() {
--- a/pkg/resource/service_scheduled_scaling.go
+++ b/pkg/resource/service_scheduled_scaling.go
@@ -599,10 +599,15 @@ func applyScheduleToState(schedule *api.AutoScalingSchedule, state *models.Servi
 
 // applyScheduleToStateWithPlan is the Create/Update counterpart of
 // applyScheduleToState. It reconciles the server response against the plan so
-// that values the user set explicitly survive server-side normalization. The
-// server drops the idle fields for a non-idle entry (idle_scaling=false), so a
-// server-only mapping would turn a planned idle_scaling=false into null and
-// trip Terraform's "produced inconsistent result after apply" check.
+// that values the user set explicitly survive server-side normalization: any
+// field the server echoes differently from how it was sent (or not at all)
+// would otherwise land in state as null and trip Terraform's "produced
+// inconsistent result after apply" check. The known instance of this was
+// UC-1252 (issue #611): the server normalized a vertical entry's equal
+// minReplicas/maxReplicas band to numReplicas and omitted the band from the
+// response. That is fixed server-side (control-plane#35956), but reconciling
+// against the plan keeps Create/Update correct regardless of which fields any
+// server version echoes.
 func applyScheduleToStateWithPlan(ctx context.Context, schedule *api.AutoScalingSchedule, planEntries types.List, state *models.ServiceScheduledScalingResourceModel) diag.Diagnostics {
 	entriesList, diags := reconcileEntriesWithPlan(ctx, schedule, planEntries)
 	if diags.HasError() {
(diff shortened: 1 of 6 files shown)


pkg/resource/service_scheduled_scaling.go
```diff
@@ -488,20 +581,91 @@ func applyScheduleToState(schedule *api.AutoScalingSchedule, state *models.Servi
 		}
 		entryValues[i] = entryModel.ObjectValue()
```
serdec: **Read rebuilds the list purely in server-response order**, with no reconciliation against prior-state order. Now that `entries` is a `List`, element identity is positional, so if the server ever returns entries in a different order than the user's config (sorted by start hour / name / id), a refresh writes state in server order and the next `plan` shows a permanent, non-converging reorder diff — the exact churn a `Set` structurally prevented.

Create/Update are safe because they write in plan order, but Read/import/refresh silently depend on an undocumented server ordering guarantee — in a PR whose whole premise is *not* trusting the response shape. Suggest reconciling here by matching on `name` (reuse the name-keyed correlation from `applyScheduleToStateWithPlan`) to preserve prior order, or at minimum documenting the hard dependency on server order preservation.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -488,20 +581,91 @@ func applyScheduleToState(schedule *api.AutoScalingSchedule, state *models.Servi
...
+// unknown).
+func applyScheduleToStateWithPlan(schedule *api.AutoScalingSchedule, planModels []models.ScheduledScalingEntryModel, state *models.ServiceScheduledScalingResourceModel) diag.Diagnostics {
+	var diags diag.Diagnostics
+
+	values := make([]attr.Value, len(planModels))
+	for i, pm := range planModels {
+		// The zero model is all-null: the fallback when the server returned no
+		// matching entry at this index.
+		var server models.ScheduledScalingEntryModel
+		if i < len(schedule.Entries) && schedule.Entries[i].Name == pm.Name.ValueString() {
```
serdec: The index+name correlation is a good defensive touch. One caveat: it assumes entry names are unique. If the server doesn't enforce uniqueness, two entries with the same name + a reordered response could still mis-correlate and fill unset fields from the wrong entry. Given the overlap-rejection this is low-likelihood, but worth a comment noting the uniqueness assumption (and it would also affect a name-based Read fix).

pkg/resource/service_scheduled_scaling.go
```diff
@@ -342,6 +346,92 @@ func (r *ServiceScheduledScalingResource) ImportState(ctx context.Context, req r
...
+		0: {
+			PriorSchema: &priorSchema,
+			StateUpgrader: func(ctx context.Context, req resource.UpgradeStateRequest, resp *resource.UpgradeStateResponse) {
+				var prior scheduledScalingModelV0
+				resp.Diagnostics.Append(req.State.Get(ctx, &prior)...)
+				if resp.Diagnostics.HasError() {
+					return
+				}
+
+				entryType := models.ScheduledScalingEntryModel{}.ObjectType()
```
serdec: Correct today since the v0 and v1 entry object types are identical. Note this couples the 0→v1 upgrade to the *current* model type: if a future schema version adds an entry field, this re-wrap breaks because v0-decoded elements won't carry it. A one-line comment noting that assumption would help the next editor.

serdec (commented):
A few inline notes from the review. The main one is the Read-path order dependency introduced by the Set→List switch; the others are minor.

serdec:
Re: the CLA check — the branch's first commit was authored by `Claude <noreply@anthropic.com>` (it was created via a claude.ai/code session), which CLA assistant counted as an unsigned committer. I've re-authored that commit to my account (content unchanged, `Co-Authored-By: Claude` trailer kept for attribution) and force-pushed. All committers are now covered by my CLA signature; the check should clear on recheck.

pkg/resource/service_scheduled_scaling.go
serdec: Fixed in 210aa0e. Read now reorders the server response to match prior-state entry order before writing state, correlated by an identity key (see the other thread — name + sorted weekdays + hours, not name alone). Entries not present in prior state append in server order, and import (empty prior state) keeps server order. Verified end-to-end against a mock that reverses entry order in every response: apply succeeds and plan converges with no reorder diff.

pkg/resource/service_scheduled_scaling.go
serdec: Fixed in 210aa0e. The correlation key is now the entry's full user-declared identity — name + sorted weekdays + start/end hours (apiEntryKey/modelEntryKey) — instead of name alone. Names may collide, but two entries cannot cover the same window (the server rejects overlaps), so the combined key is unique; the comment on the helpers documents this reasoning. The same key drives both this guard and the new Read-path reordering, and the new unit test uses two same-named entries to pin it.

pkg/resource/service_scheduled_scaling.go
serdec: Done in 210aa0e — the re-wrap now carries a comment stating it relies on the v0 and v1 entry object types being identical, and that a future entry-type change must decode and convert instead.

serdec (commented):
Follow-up review on the updated head (server-side fix control-plane#35956 assumed merged/deployed, so pre-fix refresh-diff findings are excluded). The core change is correct; remaining notes are low-severity hardening and test coverage.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -342,6 +346,95 @@ func (r *ServiceScheduledScalingResource) ImportState(ctx context.Context, req r
...
```
serdec: **Low — the upgrader couples to the *live* model entry type.** This builds the v1 target list from `models.ScheduledScalingEntryModel{}.ObjectType()` rather than a frozen v1 type. Correct today, but when a future PR bumps to v2 and changes the entry object shape, this single registered `0:` upgrader must still emit the *current* (v2) type while re-wrapping v0-typed elements — `types.ListValue` will then fail on the type mismatch. Suggest freezing an `entryObjectTypeV1()` helper (mirroring `scheduledScalingSchemaV0()`) and having the upgrader emit that, decoupling it from the live model.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -471,37 +566,169 @@ func planEntriesToAPI(ctx context.Context, entriesSet types.Set) ([]api.AutoScal
...
+
+	applyBaseConfigToState(schedule, state)
+	return diags
+}
+
+// apiEntryKey / modelEntryKey compute a schedule entry's identity from its
+// user-declared fields: name plus the window itself (sorted weekdays and
+// hours). Names alone are not guaranteed unique, but two entries cannot cover
+// the same window (the server rejects overlaps), so the combined key is.
+func apiEntryKey(e api.AutoScalingScheduleEntry) string {
```
serdec: **Very low — `apiEntryKey` and `modelEntryKey` independently `Sprintf` the same `%s|%v|%d|%d` layout.** Since these keys are compared for entry correlation, editing one format and not the other would silently break matching (entries stop correlating, degrading to null-fills / reorder handling) with no compile error. Consider a single shared `entryKey(name string, weekdays []int, start, end int)` that both call.

pkg/resource/service_scheduled_scaling.go
```diff
@@ -471,37 +566,169 @@ func planEntriesToAPI(ctx context.Context, entriesSet types.Set) ([]api.AutoScal
...
+	if diags.HasError() {
+		return diags
+	}
+	state.Entries = entriesList
+
+	applyBaseConfigToState(schedule, state)
+	return diags
+}
+
+func applyBaseConfigToState(schedule *api.AutoScalingSchedule, state *models.ServiceScheduledScalingResourceModel) {
```
serdec, line 754: **Very low (pre-existing) — `base_config` is the same failure class this PR fixes for `entries`, but has no analog.** `base_config` is `Computed + UseStateForUnknown`, so on Update its plan value is the known prior-state object; this then overwrites state from the server unconditionally. If the server ever returns a `base_config` differing from prior state during a schedule edit, `state != plan(known)` → "inconsistent result after apply" for `base_config`. Low probability (server-managed, stable across edits) and not introduced here, but worth noting the entries fix doesn't cover base_config.

pkg/resource/service_scheduled_scaling_test.go
```diff
@@ -502,3 +490,336 @@ func TestRoundTrip_NoServerNormalization(t *testing.T) {
 		t.Errorf("weekdays = %v; want [1 2 3]", got)
 	}
 }
+
+// reconcileSingleEntry drives applyScheduleToStateWithPlan for one plan entry
+// against one server-returned entry and returns the resulting state entry.
+func reconcileSingleEntry(t *testing.T, planEntry models.ScheduledScalingEntryModel, serverEntry api.AutoScalingScheduleEntry) models.ScheduledScalingEntryModel {
```
serdec, line 496: **Low — reconcile robustness isn't directly unit-tested.** `reconcileSingleEntry` only ever exercises a single, key-matching entry, but the reconcile exists precisely for the harder cases. Worth adding: (a) a multi-entry plan where the server response is **reordered**, asserting known plan fields survive and unmatched-unknown fields resolve to null without error; and (b) a plan of N entries where the server returns **N−1**, asserting `len(state) == len(plan)`.

dbashford-clickhouse:
converting this to draft while I assess if it is needed to patch an issue I may have patched in OpenAPI

dbashford-clickhouse:
Recommend we hold / not move forward with this, now that the root cause is fixed at the source.

**The fix is control-plane#35956 (UC-1252).** As this PR's description notes, once that lands the released provider round-trips vertical schedule entries with **no provider change** — the API went back to emitting an equal `minReplicas`/`maxReplicas` band for vertical entries instead of `numReplicas`. #35956 has now deployed to prod, so the `scheduled_scaling` e2e should recover on its own; we're verifying that right now on #584's e2e run.

That leaves this PR's only additive content as the `entries` Set→List hardening, and I'd pass on it:

- **It guards a class the API fix already closes.** The "does not correlate" failure was the API returning `numReplicas` (a field the provider doesn't model) → `min/max` read back as null. That's fixed server-side.
- **Set→List is backwards semantically and costs a state migration.** This resource is "recurring weekly windows where order isn't significant" — that's a set, not a list. Converting to an order-sensitive List (plus a schema version bump + `UpgradeState`) to work around Terraform's set-correlation-with-unknowns limitation trades the correct data model for a workaround, with no remaining bug to justify it.
- **It collides with #584** (horizontal autoscaling), which edits the same files (`service_scheduled_scaling.go`, the model, `api/scheduled_scaling.go`, tests, docs). Running both in flight guarantees conflicts.

If we ever want set-correlation defense-in-depth, let's evaluate it standalone after #584 merges — noting #584 already keeps its own Computed field (`autoscaling_mode`) from breaking correlation, which the post-#35956 prod e2e will confirm.

Root-cause fix tracked in UC-1252 / control-plane#35956.

PR #613, opened 2026-07-14
refactor: restructure into internal/service groups with a ServicePackage registry

jordan-simonovski:
Phase 1 of **HDX-4777** / `docs/rfcs/0001`. A behavior-preserving restructure. **no resource/data-source type-name, schema, or state changes**; registry address `registry.terraform.io/ClickHouse/clickhouse` and module path unchanged.

## What changed

- Move `pkg/` → `internal/`; split resources/data sources into `internal/service/<group>/` (`clickhouse`, `postgres`).
- Introduce the `service.ServicePackage` interface + a central `internal/service/registry` subpackage (kept as a subpackage to avoid an import cycle). The provider now composes `Resources()`/`DataSources()` by iterating registered groups instead of `GetResourceFactories()` + a hand-written data-source list.
- Inject the API client via a `service.ProviderData` container (an extensibility seam for Phase 3's second client) instead of a bare `api.Client` in `ResourceData`/`DataSourceData`. All 22 resource/data-source `Configure` methods updated accordingly.
- Delete the two build-tagged `register_{stable,debug}.go` files — their bodies were identical, so the `alpha` build path was already a no-op; `-tags alpha` still builds.
- Move all non-Go references with the code: goreleaser ldflags, composite-action (e2e/import) ldflags, Makefile `mock` target, minimock `//go:generate`, `.golangci.yml` path regexes, `.air.toml`, and `release-notes.yaml` paths.

## Verification (all green)

- `go build ./...` and `go build -tags alpha ./...`
- `go test ./...`, `go vet ./...`
- `make fmt` → `0 issues`
- `make docs` and `make docs-alpha` → no diff
- `make mock` → no diff
- No `terraform-provider-clickhouse/pkg` import stragglers remain.

Behavior-preservation independently verified beyond the build:

- **Full provider schema diff (`main` vs this branch) is byte-identical** across all 14 resources, 8 data sources, and the provider config block.
- **Live smoke test** against ClickHouse Cloud: the branch-built provider authenticates and reads a data source end-to-end through the new `registry → Configure → ProviderData → api.Client` path.
- A new `registry_test.go` assertion pins the registered surface to exactly **14 resources / 8 data sources**, so an accidentally dropped factory fails CI.

## Notable deviations from the plan (all mechanical, no redesign)

- `postgres_service_constants.go` (not in the plan's move list) moved with the postgres group — it holds postgres-only unexported consts used solely by `postgres_service.go`.
- The `Configure` rewrite covered 3 files the plan's grep didn't match: `clickpipe.go` (parameter named `request`, not `req`) and the two `clickpipe_reverse_private_endpoint*` files whose `client` field is the **concrete** `*api.ClientImpl` — those recover it via `providerData.API.(*api.ClientImpl)` to preserve behavior.

## Deferred (noted in code)

Extracting a shared `service` helper for the repeated `ProviderData` unwrap is deliberately deferred — Phase 1 stays mechanical, and Phase 3 re-touches every `Configure`. This is documented on the `service.ProviderData` type.

## Validation harness (intentionally not in this PR)

A manual `dev_overrides`-based smoke harness — build the provider from the branch, `terraform validate`/`plan` against real Cloud, plus a `main`-vs-branch schema-equality diff covering every resource — was used to validate this change. It's kept local for our own validation and left out of this PR for now.

resolves HDX-4777


jordan-simonovski:
## DSL-neutrality validation (no user-facing Terraform changes)

This PR is a behaviour-preserving restructure. To confirm it adds **nothing** to the user-facing Terraform DSL, in particular **no new provider variable configurations** (those, e.g. `clickstack_endpoint` / `clickstack_api_key`, come in a later phase). I compared the provider schema built from this branch against `origin/main`.

**Provider config variables. Identical (`origin/main` vs this branch):**

```
api_url
organization_id
timeout_seconds
token_key
token_secret
```

No provider variables added, removed, or changed. Extracted via
`terraform providers schema -json` → `.provider_schemas[].provider.block.attributes`
from a `main`-built and a branch-built provider binary; the diff is empty.

**Whole provider surface,  identical:** the same comparison across the full schema (all 14 resources, 8 data sources, and the provider config block) is byte-identical between `origin/main` and this branch.

**This is already enforced by CI.** The provider's config variables are rendered into `docs/index.md` (under `## Schema`), which `check_docs` regenerates and diffs (`git diff --exit-code`). A new provider attribute would change that file and fail the check and `check_docs` is green here. When the real DSL additions land in a later phase, they'll surface as a reviewable `docs/` diff at that time.

nunoadrego (approved):

EvandroLG (approved):

iskakaushik (approved):

PR #618, opened 2026-07-15
fix(clickpipes): don't re-send source connection on a table_mappings-only edit

### LINKED ISSUE #617: clickhouse_clickpipe: CDC table_mappings-only edit fails with AuthenticationFailed (update re-sends uri but nils unchanged credentials)
## Summary

Editing only `table_mappings` on a running CDC ClickPipe (Postgres, MySQL, or MongoDB) fails with a source-connection **authentication** error, even when the pipe's stored credentials are valid and it is actively replicating:

```
Error Updating ClickPipe
Could not update ClickPipe, unexpected error: status: 400, body:
{"error":"BAD_REQUEST: failed permissions validation: 'connectionStatus' failed:
connection() error occurred during connection handshake: auth error: sasl
conversation error: unable to authenticate using mechanism \"SCRAM-SHA-1\":
(AuthenticationFailed) Authentication failed.","status":400}
```

## Root cause

On update, `extractSourceFromPlan` always populates the source `URI`, but the update path nils the credentials when they are unchanged:

```go
source := c.extractSourceFromPlan(...)         // URI is always set
if !credentialsChanged && source.MongoDB != nil {
    source.MongoDB.Credentials = nil           // credentials dropped
}
clickPipeUpdate.Source = source                // payload: uri present, credentials absent
```

The resulting PATCH carries the connection `uri` but no `credentials`. The control plane re-validates the source connection whenever the `uri` is present, and with no credentials that handshake authenticates with nothing and fails `connectionStatus` / `AuthenticationFailed`. The pipe's stored credentials are never consulted.

This affects all three CDC source types. The nil-credentials logic exists for `Postgres`, `MySQL`, and `MongoDB` (pkg/resource/clickpipe.go, the three `if !credentialsChanged && source.X != nil` branches).

`credentialsObjectChanged` skips `password_wo`, so write-only-credential configs also resolve to `credentialsChanged == false` on a mappings-only change and hit this too. The only configs that escape are those whose apply also changes the username or a stateful password.

## Why it surfaces now

Before the pause-before-edit change (3.18.x), a `table_mappings` edit on a *running* CDC pipe failed earlier with `ClickPipe must be paused to edit` and never reached connection validation, so this was masked. Now that the provider pauses the pipe and proceeds to the edit, the update reaches `connectionStatus` validation and trips over the missing credentials. (The "pause the pipe by hand, then apply" workaround on older versions hits the same failure.)

## Reproduction

1. Create a running MongoDB (or Postgres/MySQL) CDC ClickPipe with valid `credentials { username, password }`. Let it replicate.
2. Change **only** `table_mappings` (add or remove one table). Do not touch credentials.
3. `terraform apply`.

Result: the `400 ... connectionStatus ... AuthenticationFailed` above, despite the pipe replicating fine on its stored credentials.

## Suggested fix

When an update changes only `table_mappings` (connection unchanged), send just the mapping deltas and omit the connection fields, mirroring the incremental-mapping API shape. Because `uri` / `credentials` / `settings` are all `omitempty`, clearing them on a mappings-only edit produces `{"source":{"mongodb":{"tableMappingsToAdd":[...]}}}`, which does not trigger connection re-validation. (Directly editing `table_mappings` via the API with only `tableMappingsToAdd`/`tableMappingsToRemove` works today for exactly this reason.)

More generally: the `uri` should only be re-sent when it actually changed, and whenever the `uri` is sent the credentials must be sent with it, so the connection re-validation has something to authenticate with.

Happy to open a PR along these lines.

## Environment

- Provider: 3.18.1
- Source type: MongoDB CDC (same code path for Postgres/MySQL)

david-cove:
## Problem

Editing only `table_mappings` on a running CDC ClickPipe (Postgres, MySQL, or MongoDB) fails with a source-connection authentication error, even when the pipe is replicating fine on valid stored credentials:

```
Error Updating ClickPipe ... status: 400 ... 'connectionStatus' failed:
... (AuthenticationFailed) Authentication failed.
```

`Update()` builds the source via `extractSourceFromPlan` (which always populates the `uri`) and then nils the credentials when they are unchanged. The resulting PATCH carries the connection `uri` but no `credentials`, so the control plane re-validates the source connection with nothing to authenticate with, and fails. The pipe's stored credentials are never consulted. This affects all three CDC source types, and `credentialsObjectChanged` skips `password_wo`, so write-only configs hit it too. See #617 for the full analysis.

It surfaces now because the pause-before-edit change lets the update reach connection validation; previously the edit failed earlier with `must be paused to edit`.

## Fix

When an update changes **only** `table_mappings` (connection, settings, and credentials all unchanged), PATCH just the mapping deltas and omit the connection fields. Because `uri` / `credentials` / `settings` are `omitempty`, this produces `{"source":{"<db>":{"tableMappingsToAdd":[...]}}}`, which does not trigger connection re-validation. This mirrors editing `table_mappings` directly via the API with only `tableMappingsToAdd`/`tableMappingsToRemove`, which works today.

Updates that also change the connection/settings/credentials are unchanged: they still send the full source (with credentials when those changed).

## Testing

- New unit test `TestClickPipeResource_Update_MappingsOnlyOmitsConnection_Issue617` drives `Update()` end to end through the minimock client and asserts the PATCH payload for a mappings-only edit omits `host`/`port`/`database`/`credentials`/`settings` and carries only the mapping delta.
- Existing pause-before-edit tests, the full `internal/service/clickhouse/resource` and `internal/api` unit suites, `go build ./...`, and `make fmt` (golangci-lint) all pass.

Fixes #617

### DIFF
--- a/internal/service/clickhouse/resource/clickpipe.go
+++ b/internal/service/clickhouse/resource/clickpipe.go
@@ -5262,11 +5262,22 @@ func (c *ClickPipeResource) Update(ctx context.Context, req resource.UpdateReque
 						}
 					}
 
-					if len(tableMappingsToAdd) > 0 {
-						source.Postgres.TableMappingsToAdd = tableMappingsToAdd
-					}
-					if len(tableMappingsToRemove) > 0 {
-						source.Postgres.TableMappingsToRemove = tableMappingsToRemove
+					if otherFieldsChanged {
+						if len(tableMappingsToAdd) > 0 {
+							source.Postgres.TableMappingsToAdd = tableMappingsToAdd
+						}
+						if len(tableMappingsToRemove) > 0 {
+							source.Postgres.TableMappingsToRemove = tableMappingsToRemove
+						}
+					} else {
+						// Only table_mappings changed: PATCH just the mapping deltas.
+						// Re-sending the connection (host/port/settings) makes the
+						// control plane re-validate the source connection, which fails
+						// because the unchanged credentials are omitted (#617).
+						source.Postgres = &api.ClickPipePostgresSource{
+							TableMappingsToAdd:    tableMappingsToAdd,
+							TableMappingsToRemove: tableMappingsToRemove,
+						}
 					}
 				}
 
@@ -5332,11 +5343,22 @@ func (c *ClickPipeResource) Update(ctx context.Context, req resource.UpdateReque
 						}
 					}
 
-					if len(tableMappingsToAdd) > 0 {
-						source.MySQL.TableMappingsToAdd = tableMappingsToAdd
-					}
-					if len(tableMappingsToRemove) > 0 {
-						source.MySQL.TableMappingsToRemove = tableMappingsToRemove
+					if otherFieldsChanged {
+						if len(tableMappingsToAdd) > 0 {
+							source.MySQL.TableMappingsToAdd = tableMappingsToAdd
+						}
+						if len(tableMappingsToRemove) > 0 {
+							source.MySQL.TableMappingsToRemove = tableMappingsToRemove
+						}
+					} else {
+						// Only table_mappings changed: PATCH just the mapping deltas.
+						// Re-sending the connection (host/port/settings) makes the
+						// control plane re-validate the source connection, which fails
+						// because the unchanged credentials are omitted (#617).
+						source.MySQL = &api.ClickPipeMySQLSource{
+							TableMappingsToAdd:    tableMappingsToAdd,
+							TableMappingsToRemove: tableMappingsToRemove,
+						}
 					}
 				}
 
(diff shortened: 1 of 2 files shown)


david-cove:
@tpanetti heads up, this is a follow-up to the pause-before-edit change shipped in v3.18.1. That fix is what let the update reach source validation, which surfaced this pre-existing credential-omission path. This PR has a proposed fix plus an end-to-end test if it's useful.

tpanetti (changes requested):
Awesome work david! Pulled the change and verified it works with table mappings. Let's address the otherFieldsChanged issue copilot has pointed out and I'll approve, merge, and build a new revision. Thanks for the contribution!! 🚀

david-cove:
I was just talking to a coworker and learned that the ClickHouse engineering team has root-caused a second, related control-plane bug: 

on MongoDB ClickPipes a mapping-only edit that omits credentials re-encrypts the stored credential blob on every call, growing it until it crosses an internal size limit, after which every credentials-less edit fails permanently with a 500 "Internal error" (~6 edits from clean). The validated fix is to include the credentials (source.mongodb.credentials) on every mapping edit: it replaces the blob cleanly and the edit succeeds. 

This comes from support case #48249.

My reading of this is that my proposed fix actually does the opposite of what is needed and that I should revise this to include credentials on a table_mappings-only edit. Do you agree?

tpanetti:
> My reading of this is that my proposed fix actually does the opposite of what is needed and that I should revise this to include credentials on a table_mappings-only edit. Do you agree?

That's okay, thanks for the lookout. That is a bug on our API side with a fix we have in progress that should be deployed soon, so it should be a non-issue for this PR. Let's just address the copilot comments about the partial editing drops and we should be good to go!

david-cove:
Addressed the partial-editing-drops: `otherFieldsChanged` now compares every source attribute except `table_mappings` via a shared helper, instead of an enumerated subset, so a host/authentication/tls_host/etc. change alongside a mapping edit can't be treated as mappings-only and dropped. Added a test covering that case plus the existing mappings-only-omits-connection test. Rebased onto v3.19.0; build, unit tests, and golangci-lint all green.

tpanetti (approved):

PR #623, opened 2026-07-16
feat(clickstack): add ClickStack service group

jordan-simonovski:
Phase 3 (final) of [`docs/rfcs/0001-consolidate-providers-and-service-groups.md`](docs/rfcs/0001-consolidate-providers-and-service-groups.md). Ports the entire ClickStack (HyperDX) provider into `internal/service/clickstack/` as **`clickhouse_clickstack_*`** resources and data sources.

## What's included

- **API client** — `internal/service/clickstack/client/` (verbatim port of the upstream stdlib-only client + its httptest-based tests).
- **9 resources**: `connection`, `dashboard`, `source`, `role`, `team`, `team_member`, `webhook`, `saved_search`, `alert` — all prefixed `clickhouse_clickstack_*`.
- **2 data sources**: `dashboard`, `role`.
- **ServicePackage + registry** — `clickstack.ServicePackage()` (`StabilityAlpha`) registered in `internal/service/registry`; the registry uniqueness test proves no type-name collisions (the `clickhouse_role` vs clickstack `role` clash is avoided by the prefix).
- **Provider credentials** — new optional `clickstack_endpoint` / `clickstack_api_key` attributes (+ `CLICKSTACK_ENDPOINT` / `CLICKSTACK_API_KEY` env fallbacks); `service.ProviderData` gains a `ClickStack` client field.
- **Examples + generated docs** for all 11 types; provider README gains a ClickStack section with the aliased-provider pattern. (The upstream `example` scaffold was intentionally dropped.)
- **Acceptance tests** ported into an external `clickstack_test` package (breaks the registry→clickstack cycle).

## Alpha status
ClickStack is **alpha**, expressed at runtime via `utils.AlphaWarning` on every resource + `StabilityAlpha` group metadata. Per RFC decision #4 there is **no build-tag gating** — it compiles into both stable and alpha binaries; the soft release goes out via the alpha pipeline by convention.

## Backward compatibility
No impact on existing users. Cloud/postgres resource code is untouched; the provider's cloud credential path is behavior-preserving (a bare/misconfigured block emits the same errors; full cloud creds build the same client). New provider attributes are optional and not stored in state; all existing resource/data-source **type names are unchanged**, so `terraform plan` against existing state shows no diff. A ClickStack-only provider block is now also valid.

## Verification
- `go build ./...`, `go build -tags alpha ./...`, `go test ./...` — clean
- `make lint`, `make sec`, `make cover`, `make docs-check`, `make goreleaser-check` — all green
- Registry uniqueness test passes (23 resources / 10 data sources)
- Two-pass code review (credential-gating correctness + transform integrity) — no defects
- **Acceptance tests were compile-checked only** — no live ClickStack instance was available to run the `TestAcc*` suite against.


teeohhem (approved):
Awesome stuff 👍 

Please note that the entirety of this PR was reviewed as it was developed in a private repo before moving here.

nunoadrego (approved):

PR #633, opened 2026-07-21
feat(clickstack)!: support ClickStack on ClickHouse Cloud via the Cloud API

jordan-simonovski:
## What & why

ClickStack on ClickHouse Cloud is served through the ClickHouse Cloud API (`/v1/organizations/{org}/services/{svc}/clickstack/*`, HTTP basic auth with the Cloud key/secret), **not** the self-hosted bearer-token API. The provider previously assumed the self-hosted flow worked against Cloud via a default `hyperdx-api.clickhouse.cloud` endpoint, which never actually worked. This PR adds first-class Cloud support and drops that wrong assumption.

Verified end-to-end against a live Cloud service: source import, dashboard create/update/delete, and the credential-resolution paths.

## What changed

**New Cloud auth mode.** A new provider attribute `clickstack_service_id` (or `CLICKSTACK_SERVICE_ID`) routes `clickhouse_clickstack_*` resources through the Cloud API, authenticating with the existing `organization_id` / `token_key` / `token_secret`. Self-hosted deployments continue to use `clickstack_endpoint` + `clickstack_api_key`.

**Client Cloud mode** (`internal/service/clickstack/client`): basic auth, org/service-scoped paths, `{result}` → `{data}` envelope translation, `error`-keyed error bodies, and a route-vs-resource 404 distinction so a routing 404 (a path the Cloud API does not serve yet) is never mistaken for "resource deleted" and does not silently evict state. `GetSource` falls back to list+filter (no GET-by-id on Cloud). Endpoint coverage is intentionally **not** gated client-side — the Cloud API is still rolling out the full ClickStack surface, so newly exposed endpoints work without a provider release.

**Dashboard filter IDs.** The Cloud API rejects a filter `id` on create but requires one on every filter on update. `mergeFilterIDs` carries the server-assigned id forward for existing filters and mints a placeholder for new/renamed ones (the server reassigns its own id, so the value only has to satisfy validation). Generalized from the existing `mergeTileIDs` logic.

**Credential resolution — written config wins.** When the provider block explicitly selects a mode, stray `CLICKSTACK_*` environment variables for the other mode are ignored, so a leftover variable can't manufacture a false "ambiguous configuration" abort. A mode counts as selected only when its attribute is non-null **and** non-empty (handles `attr = var.x` with an empty-defaulting var).

## Breaking change

`clickstack_api_key` now requires `clickstack_endpoint` — the `hyperdx-api.clickhouse.cloud` default is removed (it embodied the wrong assumption). Existing self-hosted configs that set an explicit endpoint are unaffected. There is no working installed base of the old Cloud-via-api-key flow, since it never functioned.

## Alpha status

`clickhouse_clickstack_*` remains alpha (runtime warning at plan/apply). Docs regenerated. Not yet exposed on Cloud: connections, saved searches, roles — these surface a route-not-found error until the Cloud API adds them.

## Testing

- Unit tests for Cloud client behavior (auth shape, envelope rewrap, 404 discrimination, unsupported-route handling), filter-id carry-forward/mint, and credential precedence (first `internal/provider` tests).
- Live Cloud API verification of the create/update/delete and source-import round trips.
- `make lint` clean; full `go test ./...` green.


docs/index.md
```diff
@@ -30,20 +30,31 @@ description: |-
...
   
-  provider "clickhouse" { # ClickHouse Cloud
+  provider "clickhouse" {
+    organization_id       = var.organization_id
+    token_key             = var.token_key
+    token_secret          = var.token_secret
+    clickstack_service_id = var.clickstack_service_id
+  }
+  
+  On ClickHouse Cloud, ClickStack manages connections, sources, dashboards, alerts, saved searches and webhooks. Sources are read-only — the platform provisions them, so an imported source can be read but not updated or destroyed (use terraform state rm to detach one). Roles, teams and team membership are managed through ClickHouse Cloud, not ClickStack — use the clickhouse_role and clickhouse_role_assignment resources — so the clickhouse_clickstack_role, clickhouse_clickstack_team and clickhouse_clickstack_team_member resources (and the clickstack_role data source) are for self-hosted ClickStack only. The team attribute on other resources is likewise not applicable on Cloud — a service is a single ClickStack team — and is rejected. Capability checks are server-side: an endpoint the Cloud API does not serve returns a route-not-found error, and newly exposed endpoints work without a provider upgrade.
```
teeohhem: >  Sources are read-only — the platform provisions them, so an imported source can be read but not updated or destroyed

Connections, not sources :)

docs/index.md
```diff
@@ -99,24 +110,39 @@ then a manual process is required after the upgrade. Please visit [https://githu
...
+```hcl
+provider "clickhouse" {
+  organization_id       = var.organization_id
+  token_key             = var.token_key
+  token_secret          = var.token_secret
+  clickstack_service_id = var.clickstack_service_id
+}
+```
+
+On ClickHouse Cloud, ClickStack manages connections, sources, dashboards, alerts, saved searches and webhooks. Sources are read-only — the platform provisions them, so an imported source can be read but not updated or destroyed (use `terraform state rm` to detach one). Roles, teams and team membership are managed through ClickHouse Cloud, not ClickStack — use the `clickhouse_role` and `clickhouse_role_assignment` resources — so the `clickhouse_clickstack_role`, `clickhouse_clickstack_team` and `clickhouse_clickstack_team_member` resources (and the `clickstack_role` data source) are for self-hosted ClickStack only. The `team` attribute on other resources is likewise not applicable on Cloud — a service is a single ClickStack team — and is rejected. Capability checks are server-side: an endpoint the Cloud API does not serve returns a route-not-found error, and newly exposed endpoints work without a provider upgrade.
```
teeohhem: Same here. Connections, not sources

docs/index.md
jordan-simonovski: This is what happens when I'm half awake at the airport...

teeohhem (approved):
LGTM!

nunoadrego (approved):

PR #644, opened 2026-07-31
fix(clickpipes): treat Kafka schema registry and immutable source fields as immutable

tanya-whyte-clickhouse:
Importing a Kafka pipe with a schema registry, then applying, failed with a 400 due to a validation error. This PR fixes the provider-side contributors to this bug.

**Changes:**
- Schema registry is immutable, so changing it now forces replacement.
  - The first plan post-import has no credentials in state, so it does not force replacement here.
  - Unknown-at-plan-time values do not force replacement, but warn.
- Stop sending immutable Kafka fields on update. Mutable fields are still sent when they change.
- Clarify documentation and messages related to Kafka pipe immutable fields.

See [CLP-1103](https://linear.app/clickhouse/issue/CLP-1103/bug-in-terraform-provider-v3173-preventing-clickpipe-updates-with) for companion OpenAPI PR and testing.


tpanetti (approved):
LGTM. Kafka E2E is passing so should be good for existing.

Issue #558, opened 2026-06-01
clickhouse_clickpipe: spurious "Cannot modify excluded_columns" on plan/refresh (excluded_columns modeled as ordered List)


richard-artisan:
### Description

`excluded_columns` on a Postgres CDC `clickhouse_clickpipe` table mapping is modeled as an **ordered** `types.List`, and the `ModifyPlan` "table mappings cannot be modified" guard compares prior-state against planned config using order-sensitive `List.Equal`. Because the ClickHouse API does not guarantee the order in which excluded columns are returned, a `Read`/refresh can rewrite the stored order, after which every subsequent `plan` fails with:

```
Cannot modify excluded_columns for existing table mapping '<schema>.<table>'. Table mappings cannot be updated - you must remove the old mapping and add a new one.
```

even though the *set* of excluded columns is unchanged. The pipe then cannot be planned, applied, or refreshed without manual state surgery or removing and re-adding the mapping.

> Note on context: we consume this provider through the Pulumi Terraform bridge (`@pulumi/clickhouse` 3.14.0), where it surfaces as a hard failure on `pulumi refresh`. The defect itself is in the provider's framework-level plan logic (ordered `List` plus order-sensitive comparison) and is CLI-agnostic. It reproduces with plain Terraform whenever the API returns `excluded_columns` in a different order than the configuration (see Actual behavior).

### Cloud provider

AWS

### Service tier

Development (this is a provider-side plan bug and is tier-independent)

### Affected resource(s)

`clickhouse_clickpipe` (Postgres CDC source). The same ordered-`List` plus order-sensitive `Equal` pattern is present for the MySQL and MongoDB sources.

### Provider version

`3.14.0` (current latest stable). The relevant logic is verified unchanged through `3.15.0-alpha3` / `main` (`8acb242`).

### Terraform version

N/A. Consumed via the Pulumi Terraform bridge (pulumi-terraform-bridge), provider parameterization `terraform-provider` 1.1.1. The defect is independent of the driving CLI (see Description).

### Terraform configuration

```hcl
resource "clickhouse_clickpipe" "cdc" {
  service_id = clickhouse_service.analytics.id
  name       = "aurora-cdc"

  source = {
    postgres = {
      # ...host / credentials / settings...
      table_mappings = [
        {
          source_schema = "public"
          source_table  = "lead"
          target_table  = "lead"
          table_engine  = "ReplacingMergeTree"
[...]

PR #649, opened 2026-08-03
feat(clickpipes): add partition_by_expr to Postgres CDC table_mappings

### LINKED ISSUE #648: partition_by_expr missing from table_mappings for Postgres CDC ClickPipes (API supports it, Terraform provider doesn't)
### Description

The ClickHouse Cloud ClickPipes API supports a `partitionByExpr` field on Postgres CDC table mappings — a real ClickHouse `PARTITION BY` expression applied to the destination table, distinct from `partitionKey` (the source-side snapshot-parallelism hint). It's documented in the [create-clickpipe](https://clickhouse.com/docs/products/cloud/api-reference/clickpipes/create-clickpipe) and [get-clickpipe](https://clickhouse.com/docs/products/cloud/api-reference/clickpipes/get-clickpipe) API reference, and it's exposed in the Cloud console UI (step 8 of [the Postgres CDC setup guide](https://clickhouse.com/docs/integrations/clickpipes/postgres): "you can provide a custom `PARTITION BY` expression to control destination ClickHouse table partitioning").

The Terraform provider's `table_mappings` schema (checked through v3.22.1) only exposes `use_custom_sorting_key`, `sorting_keys`, `partition_key`, `table_engine`, `target_table`, `source_table`, `source_schema_name`. There is no `partition_by_expr` field, so this cannot be set via Terraform.

This looks like the same class of gap tracked in [clickhousectl#308](https://github.com/ClickHouse/clickhousectl/issues/308), which explicitly lists:
- `ClickPipePostgresPipeTableMapping.partitionByExpr` — missing from models
- `ClickPipePatchPostgresPipeRemoveTableMapping.partitionByExpr` — missing from models

Related to #571 (which correctly established that `destination.table_definition.partition_by` is inert for CDC pipes and that per-table settings belong in `table_mappings`) — but `table_mappings` itself doesn't yet expose the one field that would actually let CDC pipe destination tables be partitioned.

### Expected behavior

`table_mappings` entries should support a `partition_by_expr` (or similarly named) attribute that maps to the API's `partitionByExpr`, so a Postgres CDC destination table's `PARTITION BY` can be set declaratively, the same way it can via the UI or a direct API call.

### Actual behavior

No such field exists in `table_mappings`. The only workaround today is setting the destination table engine to `Null` and building a separate materialized view + manually-defined partitioned table outside Terraform, or calling the ClickPipes API directly outside of Terraform and importing/accepting drift.

### Provider version

v3.22.1 (also confirmed missing back to v3.14.0)

### Affected resource(s)

`clickhouse_clickpipe` (Postgres CDC `source.postgres.table_mappings`)

tpanetti:
Closes #648.

## What

Exposes the API's `partitionByExpr` on Postgres CDC `table_mappings`, so the destination table's ClickHouse `PARTITION BY` expression can be set declaratively when ClickPipes creates the table. Only Postgres CDC supports this field per the control-plane OpenAPI spec — MySQL and MongoDB table mappings don't have it.

```hcl
table_mappings = [{
  source_schema_name = "public"
  source_table       = "events"
  target_table       = "events"
  partition_by_expr  = "toYYYYMM(created_at)"
}]
```

## Also fixes: ModifyPlan mapping-immutability guard gap

The plan-time guard that rejects in-place edits to existing table mappings ("Cannot modify X for existing table mapping...") had a stale field checklist: it covered `target_table`, `excluded_columns`, `sorting_keys`, and `use_custom_sorting_key`, but was never extended when `table_engine`/`partition_key` were added. Changing one of those on an existing mapping slipped past the plan, was silently omitted from the Update PATCH (the add/remove diff keys on mapping identity), and surfaced as a confusing "Provider produced inconsistent result after apply".

`table_engine`, `partition_key`, and `partition_by_expr` are now in the checklist, so these edits fail at `terraform plan` with the guard's existing remove-the-mapping-and-re-add guidance.

An in-place edit cannot be expressed against the API at all: the control plane rejects add+remove of the same source table within a single PATCH by design, because re-adding a mapping re-snapshots into the already-populated destination table. The plan-time error (rather than an apply-time 400 on an already-paused pipe) is therefore the contract. No false positives for existing configs: state sync preserves null `table_engine` when the API echoes a server default and maps empty `partition_key`/`partition_by_expr` to null, covered by the "identical value fields" subtest.

## Testing

- `TestClickPipeResource_ModifyPlan_MappingValueFieldsImmutable_Issue648`: per-field plan-time rejection subtests + a no-false-positive control.
- Existing #558 guard tests still pass (their plan-driver closures were extracted into shared helpers).
- `go build ./...`, full `go test ./...`, `gofmt`, and `make docs-alpha` drift check all clean.

## Follow-up (out of scope)

The MySQL mapping guard has the same stale-checklist gap for `table_engine`/`partition_key` — same one-line-per-field fix when picked up.

### DIFF
--- a/docs/resources/clickpipe.md
+++ b/docs/resources/clickpipe.md
@@ -558,6 +558,7 @@ Required:
 Optional:
 
 - `excluded_columns` (Set of String) Columns to exclude from replication.
+- `partition_by_expr` (String) ClickHouse PARTITION BY expression applied to the destination table when ClickPipes creates it. Cannot be changed on an existing table mapping: remove the mapping in one apply, then re-add it with the new value in a subsequent apply (re-adding re-snapshots the table).
 - `partition_key` (String) Custom partitioning column used for parallel snapshotting. Only beneficial for PostgreSQL 13 (no benefit for PG14+, which supports indexed ctid scans). Must be an indexed column of type: `smallint`, `integer`, `bigint`, `timestamp without time zone`, or `timestamp with time zone`. Unrelated to ClickHouse partitioning.
 - `sorting_keys` (List of String) Ordered list of columns to use as sorting key for the target table. Required when use_custom_sorting_key is true.
 - `table_engine` (String) Table engine to use for the target table. (`MergeTree`, `ReplacingMergeTree`, `Null`)
--- a/internal/api/clickpipe_models.go
+++ b/internal/api/clickpipe_models.go
@@ -210,6 +210,7 @@ type ClickPipePostgresTableMapping struct {
 	SortingKeys         []string `json:"sortingKeys,omitempty"`
 	TableEngine         *string  `json:"tableEngine,omitempty"`
 	PartitionKey        *string  `json:"partitionKey,omitempty"`
+	PartitionByExpr     *string  `json:"partitionByExpr,omitempty"`
 }
 
 type ClickPipeMySQLSource struct {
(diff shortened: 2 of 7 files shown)


markdawson (approved):
Looks good to me

PR #656, opened 2026-08-13
feat(clickpipes): add TLS options to Postgres and MongoDB sources

masterashu:
## Summary

- Postgres source: adds `disable_tls` and `skip_cert_verification`, mirroring the existing MySQL source options
- MongoDB source: adds `skip_cert_verification` (it already had `disable_tls`)

All new attributes are optional booleans defaulting to `false`, sent to the API as `disableTls` / `skipCertVerification`, wired through the terraform model, schema, plan→API extraction, and API→state sync.

## Testing

- `go build ./...` and `go test ./internal/...` pass
- `make docs-alpha` produces a clean diff (only the new attributes)


whites11 (approved):

docs/resources/clickpipe.md
```diff
@@ -510,8 +510,10 @@ Optional:
 
 - `authentication` (String) Authentication method for Postgres connection. Supported values: `basic`, `IAM_ROLE`. Default is `basic`.
 - `ca_certificate` (String) PEM encoded CA certificate to validate the Postgres server certificate.
+- `disable_tls` (Boolean) Disable TLS for the Postgres connection.
 - `iam_role` (String) IAM role ARN for IAM authentication. Required when authentication is set to `IAM_ROLE`.
 - `port` (Number) The port of the Postgres instance. Default is 5432.
+- `skip_cert_verification` (Boolean) Skip certificate verification for the Postgres connection.
```
pfcoperez, line 517: Given that all three connectors have the same options, MongoDB `skip_cert_verification` should be documented too. I'd expand this PR to align all parameters across all the connectors.

internal/api/clickpipe_models.go
```diff
@@ -180,6 +180,8 @@ type ClickPipePostgresSource struct {
 	IAMRole               *string                         `json:"iamRole,omitempty"`
 	TLSHost               *string                         `json:"tlsHost,omitempty"`
 	CACertificate         *string                         `json:"caCertificate,omitempty"`
+	DisableTLS            *bool                           `json:"disableTls,omitempty"`
```
pfcoperez, line 183: Same comment, let's add it for MongoDB too

pfcoperez (commented):
Changes LGTM, thanks for putting this together!
However, I'd expand them to MongoDB either on the same PR or in close-in-time follow up. 

The idea of TLS unification is that these parameters are the same for all the data sources. I think it should be easy to expand this one with MongoDB changes too.

pfcoperez (approved):

internal/service/clickhouse/resource/clickpipe.go
```diff
@@ -4863,6 +4905,12 @@ func (c *ClickPipeResource) syncClickPipeState(ctx context.Context, state *model
 			mongodbModel.DisableTLS = stateMongoDBModel.DisableTLS
 		}
 
+		if clickPipe.Source.MongoDB.SkipCertVerification != nil {
+			mongodbModel.SkipCertVerification = types.BoolValue(*clickPipe.Source.MongoDB.SkipCertVerification)
+		} else {
+			mongodbModel.SkipCertVerification = types.BoolValue(false)
+		}
```
tpanetti, line 4912: To avoid drift if the API omits the field simply set the value to the config value on the else. Just like we do above for DisableTLS
```suggestion
		} else {
			mongodbModel.SkipCertVerification = stateMongoDBModel.SkipCertVerification
		}
```

tpanetti (approved):
Minor fix, otherwise LGTM

PR #669, opened 2026-08-26
feat(clickpipes): add SSH key resource and ssh_key_resource_id pipe attach

markdawson:
## Summary

Adds Terraform support for SSH-tunneled ClickPipes (CLP-283), covering the two public API surfaces that shipped earlier:

### New resource: `clickhouse_clickpipes_ssh_key`

A standalone, service-scoped SSH key resource. The server generates an Ed25519 keypair and returns the OpenSSH public key; the private key never leaves ClickHouse Cloud.

- Full CRUD + import (`terraform import ... <service_id>:<ssh_key_id>`)
- `name`, `host`, `port`, `username` are immutable (`RequiresReplace`); only `description` updates in place
- Exposes `public_key`, `status` (`pending`/`active`/`failed`), `status_message`, and `last_validated_at` as computed attributes
- Connectivity validation is a day-2 out-of-band operation: it requires the public key to already be installed in the bastion's `authorized_keys`, so it is not run automatically on create (documented in the resource description)

### New pipe attribute: `source.*.ssh_key_resource_id`

Added to the **Kafka, Postgres, MySQL, and MongoDB** sources — the exact set the public API accepts it on. The field is create-only: `RequiresReplace` in the schema, and omitted from PATCH payloads to match the API's 400-on-update behavior. State sync falls back to prior state when the API response omits the field, avoiding phantom diffs on refresh.

## Notes

- Resource registered in the `clickhouse` ServicePackage; registry count test updated accordingly
- Client mock regenerated with minimock v3.4.7
- Docs regenerated via `make docs`; only the four `ssh_key_resource_id` entries are included (the unpinned tfplugindocs fork currently strips the `[Write-only]` annotations repo-wide, so that unrelated churn was reverted)
- `ValidateSSHKey` is included on the API client for the documented day-2 validation flow, though no resource calls it yet

## Testing

- Unit tests for extract/sync of the new field on Kafka and MongoDB, including the create-only guard (omitted on update)
- Full suite: `go build`, `go vet`, `go test ./internal/...`, and `make lint` all clean
- The underlying API path was verified end-to-end on production 2026-08-19 via the public OpenAPI (key create → pubkey install → validate → tunneled pipe running)


internal/service/clickhouse/resource/clickpipe_ssh_key_resource_id_test.go
tpanetti, line 1: three-branch state-sync fallback added at clickpipe.go:4191, 4587, 4828, and 4991 has no test, and the Postgres and MySQL extract paths are untested. One sync test asserting "API returns nil, prior state set, value preserved" plus mirrored extract tests for Postgres and MySQL would cover the remaining new branches.

internal/service/clickhouse/resource/clickpipe_ssh_key.go
```diff
@@ -0,0 +1,294 @@
...
+				Required:            true,
+				MarkdownDescription: "Name of the SSH key resource. Unique per service and immutable; changing it forces resource replacement.",
+				PlanModifiers: []planmodifier.String{
+					stringplanmodifier.RequiresReplace(),
+				},
+			},
+			"description": schema.StringAttribute{
+				Optional:            true,
+				Computed:            true,
+				MarkdownDescription: "Description of the SSH key resource. Immutable; changing it forces resource replacement (the public API has no update operation for SSH key resources).",
```
tpanetti: API caps description at 255 so anything over would throw a 400, add `stringvalidator.LengthAtMost(255)`

tpanetti (approved):
LGTM, nice compartmentalizing outside of resource/clickpipe.go.

nunoadrego (approved):

internal/service/clickhouse/resource/clickpipe_ssh_key_resource_id_test.go
markdawson: Thank you for catching. I added tests for Postgres and MySQL. I also fixed the sync test.

internal/service/clickhouse/resource/clickpipe_ssh_key.go
markdawson: Done, added `stringvalidator.LengthAtMost(255)` on description

PR #670, opened 2026-08-26
fix(clickstack): stop the e2e discovery step failing silently on curl exit 23

jordan-simonovski:
The nightly `clickstack` e2e has failed on every run since #667 merged, on all three Terraform versions. Other examples are unaffected.

## What happened

The "Discover ClickStack service and connection" step exited **23** with no output at all. 

Destroy then failed with "No value for required variable" because discovery never appended the ids, which is the error you see first and is not the actual problem.

Exit 23 is curl's write error. The probe inside the service loop was the only
unguarded call in the script, so `set -e` killed it with curl's numeric exit and
no message. It was also the only call pairing `-o /dev/null` with `--retry`.

I could not reproduce it locally: same curl invocation, same API, exit 0 against
both an onboarded and a non-onboarded service. So whether the trigger was that
option combination on the runner's curl or a retry against a transient gateway
error is unresolved. Both are gone rather than guessing between them.

## Changes

- The probe captures body and status together instead of `-o /dev/null -w`, and every caller checks curl's exit status. A failure now names the URL and the exit code.
- `--retry-all-errors` dropped. It needs curl >= 7.71 and nothing here needs retries on non-transient errors.
- Destroy skips when nothing was applied. It runs on failure now, so it can be   reached before `init`, where it fails on its own missing variables and hides  the real error. This affects every example, and is why the log showed two failures instead of one.

## Verification

Discovery, apply, import, update and destroy all pass against the live `[e2e]-clickstack` service: 6 added, import checks passed, 6 changed, 6 destroyed. The failure path was checked too — bad credentials now report `UNAUTHORIZED` rather than exiting silently.

Local verification is what missed this the first time, so this wants a manual `E2E tests` dispatch on the branch before merge.

### DIFF
--- a/.github/actions/e2e/action.yaml
+++ b/.github/actions/e2e/action.yaml
@@ -486,10 +486,23 @@ runs:
         echo "::group::Run terraform destroy"
         cd "examples/full/${{ inputs.test_name }}/${{ inputs.cloud_provider }}"
 
-        # Since destroy runs on failure too, it can now be reached before init
-        # ran or before anything was applied. Without this it fails on its own
-        # terms -- missing variables, no provider -- and buries the real error.
-        if [ ! -d .terraform ] || [ ! -s terraform.tfstate ]
+        # Since destroy runs on failure too, it can now be reached before
+        # anything was applied. Without this it fails on its own terms -- missing
+        # variables, no provider -- and buries the real error.
+        #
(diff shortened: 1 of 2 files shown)


jordan-simonovski:
Related PR for role management is incoming in the HDX repo

wrn14897 (approved):

nunoadrego (approved):

PR #676, opened 2026-08-31
feat(service): expose the API-assigned password when none is supplied

SaiPrasanna9:
## Summary

`CreateService`'s response already includes a system-assigned password whenever the create request doesn't set one (confirmed directly against the API — a raw create call with no password field returns one in `result.password`). The resource currently discards this value:

```go
s, _, err := r.client.CreateService(ctx, service)
```

This forces every caller to either supply an explicit `password`/`password_hash`, or use `password_wo = ""` (satisfies the schema's `AtLeastOneOf` validator, which only checks that the attribute it's attached to is non-null — an empty string qualifies — without triggering the follow-up `UpdateServicePassword` call, since that only fires when `len(passwordWO) > 0`) and have no way to retrieve the resulting password afterward — it's created, but effectively lost.

This matters in particular for a caller whose credentials only cover `control-plane:organization:create-service` and don't include `control-plane:service:manage` (needed for `UpdateServicePassword`) — which can't be scoped to the new service in advance since its ID doesn't exist until after creation. Skipping the explicit password lets creation succeed with a narrower permission set, but only if the resulting password is actually retrievable somewhere.

## Change

Adds a new `generated_password` (`Computed`, `Sensitive`) attribute that captures this value on create.

- Reaching this path requires `password_wo = ""` **and** `password_wo_version` set (e.g. to `1`) — omitting `password`/`password_hash`/`password_wo` entirely is rejected at plan time by `AtLeastOneOf`, and `password_wo` itself requires `password_wo_version` via `AlsoRequires` whenever it has any non-null value, including an empty string. Both documented on the attribute now.
- Stored as null (not `""`) if the create response's password happens to be empty rather than absent.
- Nulled out if an explicit `password`/`password_hash`/non-empty `password_wo` is also supplied in the same create, since that overwrites the generated one before it's ever used.
- Nulled out by `Update` too, whenever that apply issues an `UpdateServicePassword` call — otherwise `UseStateForUnknown` would carry forward a value the provider *knows* is wrong. The remaining documented caveat is therefore only about passwords changed **outside** Terraform, which genuinely can't be tracked.
- Null for secondary/hydra services (`data_warehouse_id` set), which share the parent's default user and never get their own password.
- Not a schema version bump — this adds a new independent `Computed` leaf attribute with no restructuring of existing ones. Existing state decodes it as null without needing an `UpgradeState` path.

## Testing

`TestServiceResource_Create_generatedPassword` covers: no password supplied, explicit `password`, explicit non-empty `password_wo`, explicit `password_hash`, and secondary services. `TestServiceResource_Create_generatedPassword_emptyResponseStaysNull` covers an empty (not absent) password in the create response. `TestServiceResource_Update_generatedPasswordClearedOnPasswordChange` covers both Update cases (password changed → cleared; nothing password-related changed → preserved).

Each of the five branches this PR adds was verified by mutation — removing any one of the nulling assignments fails at least one test.

`go build ./...`, `go vet ./...`, and `go test ./...` all pass locally. `make docs` failed in my environment on an unrelated issue (Terraform CLI download checksum/GPG verification), so I updated `docs/resources/service.md` by hand to match the existing format.

## Open question for maintainers

Activating this feature via `password_wo = ""` is a side effect of how the existing validators are written (non-null-but-empty satisfies `AtLeastOneOf` without triggering the update call) rather than an explicitly expressible intent. A cleaner long-term shape might be a dedicated `Optional`+`Computed` boolean (e.g. `generate_password`) added to the `AtLeastOneOf` set, making "let the API assign one" a first-class option instead of a documented workaround. Leaving that as a design decision for maintainers rather than including it here.

## Known follow-up, out of scope here

Every error path between `CreateService` and the final `resp.State.Set` (including the 90-minute `WaitForServiceState`) discards state entirely, orphaning the created service and, for this attribute specifically, losing the only copy of a password a caller without `service:manage` has no other way to recover. Worth a separate issue/PR to persist partial state after a successful create.


rndD (approved):
thanks

GMartinez-Sisti (approved):

PR #679, opened 2026-09-01
feat(clickstack): support multiple notification channels on alerts

jordan-simonovski:
The ClickStack v2 alerts API now accepts a `channels` array alongside the single `channel` field it mirrors from `channels[0]`, so one alert can notify up to ten destinations. This exposes `channels` on `clickhouse_clickstack_alert` and deprecates `channel`, which keeps working but can only ever reach one target. Exactly one of the two must be set; migration steps are in the README under "Upgrading to version >= 3.26.0".

Both attributes are modelled as framework types rather than a Go pointer and slice, which fixes a plan-time failure `channel` has carried since v3.20.0. A config that takes either of them from a module output or a data source leaves the whole attribute unknown until Terraform resolves it, and reflecting that into `*alertChannelModel` or `[]alertChannelModel` aborts the plan with "Received unknown value, however the target type cannot handle unknown values. This is always an error in the provider."

Duplicate detection now skips entries whose `webhook_id` is still unknown. An unknown `types.String` reads back as `""`, so two channels pointing at webhooks created in the same apply keyed identically and the second was rejected as a duplicate of the first — which broke both the documented example and the acceptance test on any fresh apply.

The exactly-one-of check runs on `IsNull()` before any decoding, because an attribute set to an unresolved reference is still set. Deciding it on the decoded value let a config set `channel` and `channels` together with no plan-time error, after which the write silently preferred the list and dropped `channel`.

The CI e2e alert moves to `channels` because import always populates that field and `e2e-import.sh` reimports the alert and asserts a clean plan, so leaving it on `channel` would have turned the pipeline red. A second alert there keeps the deprecated form exercised, since CI never runs the acceptance tests and that is the form every existing config uses.

Verified against the live staging API rather than only a local stub: create with both webhook ids unknown at plan time, reorder the list, downgrade to `channel` — where the server reduced the stored list from two entries to one, confirming an update replaces the channel list rather than merging into it — upgrade back, and import, with every step converging to no changes.

The same unknown-value defect exists on six nested attributes across `clickstack_source` and `clickstack_role`; those are out of scope here and tracked in HDX-5245.


nunoadrego (approved):

pulpdrew (approved):
LGTM

PR #681, opened 2026-09-02
feat(clickpipes): add Kafka Protobuf schema upload

EvandroLG:
## Why

Kafka ClickPipes support uploading a Protobuf schema directly, but the Terraform provider only exposed schema-registry configuration. This adds provider support so Kafka ClickPipes can be managed consistently through Terraform.

## What

- Adds the sensitive, replacement-only `source.kafka.protobuf_schema` attribute
- Validates format compatibility, base64 content, mutual exclusivity with schema registry, and the 1 MiB encoded limit
- Sends the schema on creation, omits it from updates, and preserves it in state because the API does not return it
- Redacts Protobuf schemas from API logs
- Adds unit coverage, resource documentation, and a `filebase64` example

## QA

1. Run `go test ./...`
2. Run `/tmp/golangci-lint run --new-from-rev=main`
3. Run `make docs-validate`
4. Run `terraform fmt -check examples/resources/clickhouse_clickpipe/resource_kafka_protobuf.tf`

## Risk

Low. The new attribute is optional and existing Kafka ClickPipe configurations are unchanged.


internal/service/clickhouse/resource/clickpipe.go
```diff
@@ -249,6 +249,14 @@ func (c *ClickPipeResource) Schema(_ context.Context, _ resource.SchemaRequest,
 									stringplanmodifier.RequiresReplace(),
 								},
 							},
+							"protobuf_schema": schema.StringAttribute{
+								MarkdownDescription: "Base64-encoded Protobuf schema used instead of `schema_registry`. Use `filebase64()` with a `.proto` or serialized `FileDescriptorSet` file up to 768 KiB. Requires `format = \"Protobuf\"` and forces replacement when changed.",
```
whites11, line 253: if it's "instead of schema_registry" should we validate users don't set both fields?

whites11: I see you did it all good

whites11 (approved):

nunoadrego (approved):

PR #683, opened 2026-09-03
feat(clickstack): manage dashboard tile alerts, and fix state handling on update 404s

jordan-simonovski:
Adds dashboard tile alerts to the ClickStack Terraform provider. Closes HDX-5278. Also fixes a state-handling bug that review turned up in four ClickStack resources, described at the bottom — it is not tile-alert specific.

**Outcome**

- `clickhouse_clickstack_alert` can alert on a dashboard tile: `source = "tile"` with `dashboard_id` and `tile_id`. Saved-search alerts are unchanged and `source` defaults to `saved_search`, so existing configs and state see no diff and no replacement after upgrade.
- `clickhouse_clickstack_dashboard` exposes a computed `tile_ids` map (tile name to server-assigned id). Alerts reference `tile_id = clickhouse_clickstack_dashboard.x.tile_ids["<tile name>"]`.
- Verified live against ClickStack Cloud: create, clean re-plan, and a `-generate-config-out` import round-trip.

**Why `tile_ids`**

- The server assigns tile ids. An authored id is kept only if the dashboard already has it, so the old "pin an id in dashboard_json" guidance never worked. Live testing failed with `Tile not found` and led to this design.

**Changes**

- Alert resource: `source`, `dashboard_id`, `tile_id` attributes; plan-time validation for exactly one target per source, no empty-string ids, and no `group_by` on tile alerts (the API drops it). `source` replace rule ignores a null prior value. `tile_id` only forces replacement on a known change, because `tile_ids` goes unknown on every dashboard body edit and the id resolves back to the same value. `dashboard_id` forces replacement plainly: the dashboard's id carries `UseStateForUnknown`, so it only goes unknown when the dashboard itself is being replaced, and the server deletes that dashboard's tile alerts along with it, so the alert cannot outlive it.
- Alerts client: sends `source`, `dashboardId`, `tileId`; defaults `source` only when unset.
- `Read` refuses alerts whose source the provider does not model (the API is gaining `inline`, see hyperdxio/hyperdx#3043), so importing one fails clearly instead of planning a replacement.
- Dashboard resource: `tile_ids` planned by running the same carry-forward the update path uses, so unchanged tiles stay known and UI drift stays undetected as documented.
- Tile-id carry-forward: unique names match before position; authored ids the server does not know are dropped. This also fixes a pre-existing bug where a dashboard JSON with unknown ids had every tile id re-minted on each update.
- Docs and examples updated for the two-step import (dashboard, then each alert) and the hand edits a generated import config needs.
- Tests: unit coverage for validation, round-trip, the plan modifiers and the merge; `TestAccAlertResource_Tile`; the Cloud e2e example gains a number tile with an alert re-imported by `e2e-import.sh`.

**Update no longer clears state on a 404**

Reviewing the replacement rule turned up that `alertResource.Update` cleared state when the server returned a 404, and so did dashboard, saved search and webhook. The framework's check is `if !resp.Diagnostics.HasError() && updateResp.State.Raw.Equal(nullSchemaData)`, and `RemoveResource` nulls state without setting a diagnostic, so it fired every time: anyone whose resource was deleted outside Terraform got "Missing Resource State After Update" instead of being told what happened. All four now error with the id and what to do, and prior state survives an errored Update so the next refresh converges through Read, where removing the resource is legal. The dashboard's comment claimed the old behaviour was deliberate, "drop it from state so the next plan recreates it, rather than hard-erroring", but that outcome is not reachable from Update.

The alert path is reachable without anyone touching the alert: the server deletes a tile alert once its tile stops being alertable, and the dashboard's tile id stays stable across that edit, so `tile_id` shows no diff and the alert is planned as a plain in-place update. The other three are ordinary out-of-band deletions. Dashboard, saved search and webhook shipped in v3.26.0, so this does change released behaviour, but only the message on a path that already failed every time.

The tests that pinned the old behaviour asserted nothing: `resp` was built with no `Raw`, which is already null, so "expected resource removed from state" passed whether or not `RemoveResource` ran. Alert and dashboard now seed prior state and assert the field the plan does not carry survives; saved search and webhook had no Update-404 test at all and get one each.

**Behaviour to know**

- Alerted tiles need a unique, stable name. Renaming one mints a new id and the plan fails with an invalid `tile_ids` index until the reference is updated.
- The server deletes a tile alert when its tile is removed or changed to a non-alertable type; Terraform then plans to recreate it, which fails with `Tile not found` until the tile is restored — and an update that reaches the deleted alert first now fails with a clear error rather than the framework's.
- Under `-refresh=false` the first plan after upgrade shows a one-time in-place fill of `source` and `tile_ids`.

Follow-up: embedding alert config inside `dashboard_json` is HDX-5279.


internal/service/clickstack/alert_resource.go
```diff
@@ -59,14 +61,74 @@ const channelTypeWebhook = "webhook"
...
+	return stringplanmodifier.RequiresReplaceIf(
+		func(_ context.Context, req planmodifier.StringRequest, resp *stringplanmodifier.RequiresReplaceIfFuncResponse) {
+			resp.RequiresReplace = !req.PlanValue.IsUnknown()
+		},
+		"Changing "+attrName+" forces replacement; an unknown planned value does not, the apply sends the resolved id in place.",
+		"Changing `"+attrName+"` forces replacement; an unknown planned value does not, the apply sends the resolved id in place.",
+	)
+}
+
 func isRangeThresholdType(t string) bool { return slices.Contains(alertRangeThresholdTypes, t) }
```
wrn14897, line 104: `targetRequiresReplace` skips replacement when the planned value is unknown, and is applied to both tile_id and dashboard_id. Since the dashboard id uses UseStateForUnknown (dashboard_resource.go:83), dashboard_id is only unknown during an update when the dashboard itself is being replaced. In that scenario the server cascade-deletes the tile alerts, the alert is planned as an in-place update, Update gets ErrNotFound and calls RemoveResource, and the framework fails with "Missing Resource State After Update". Suggested fix: use plain RequiresReplace() (or the same null-prior exemption as sourceRequiresReplace) for dashboard_id; keep the unknown exemption only on tile_id. Update the dashboard_id docs accordingly.

jordan-simonovski: Yep, good call. I've found this in a few other resources too, so I'll apply the fix for everything as part of this PR

wrn14897 (approved):

PR #695, opened 2026-09-09
feat(clickhouse): add Query API endpoint resource

francisconeves-clickhouse:
## Summary

Adds beta resources for managing Query API Endpoints through terraform

API docs: https://api.clickhouse.cloud/v1

## Notes

- `clickhouse_query_api_endpoint` is in beta and may change in future releases.


docs/resources/query_api_endpoint.md
```diff
@@ -0,0 +1,85 @@
+---
```
GMartinez-Sisti, line 1: Maybe this file could have the exact name as the resource `clickhouse_query_api_endpoint`? This will make it easier to identify and configure ownership from filename matching.

francisconeves-clickhouse: Thanks for taking a look! 

I think this file is generated by `tfplugindocs` which removes the provider prefix. I think the other resources don't have the prefix as well right? 🙏

GMartinez-Sisti: Thanks for checking, I'll take a note as a follow up then 🙏

GMartinez-Sisti (approved):

rndD (approved):

PR #699, opened 2026-09-11
feat(clickhouse): support custom instance profile and profile discovery data source

luisliuchao:
This PR lets BYOC organizations create services with a custom instance profile via Terraform, and discover which profiles are available without contacting support.

## `profile` attribute on `clickhouse_service`

- New `profile` attribute (Optional + Computed). The Cloud API only accepts `profile` on service creation (`POST /services`) — PATCH does not support it — so the attribute uses `RequiresReplace`: changing it in config shows `# forces replacement` in the plan.
- `UseStateForUnknown` plus Computed prevents drift/perma-diffs for services that get a profile assigned server-side (GET always returns it).
- Create passes the planned profile to the API; `syncServiceState` refreshes it on read/import; the V0→V1 state upgrader nulls the new field so the next refresh fills it.
- `profile` is also exposed as a computed read-only field on the `clickhouse_service` and `clickhouse_services` data sources.

## New `clickhouse_service_profiles` data source

Wraps `GET /v1/organizations/{organizationId}/serviceProfiles` so users can self-discover available profiles. Schema: `region_id` (required — profiles are configured per region), `byoc_id` (optional — BYOC profiles are only returned when set), computed `profiles[]` with `profile`, `cpu_cores`, `memory_gi`. Returns a known empty list (not null) when no custom profiles are available (e.g. organizations without ENTERPRISE custom profiles or BYOC infrastructure).

The documented usage is a **two-step workflow**:

**Step 1 — discover.** Apply the data source alone and inspect the output to pick a profile:

```hcl
data "clickhouse_service_profiles" "byoc" {
  region_id = "us-east-1"
  byoc_id   = var.byoc_id # optional; BYOC profiles are only returned when set
}

output "available_profiles" {
  value = data.clickhouse_service_profiles.byoc.profiles
}
```

**Step 2 — pin.** Hardcode the chosen profile name in the service config. The profile name is the only unique key (multiple profiles can share the same `memory_gi`), so the example deliberately avoids dynamic selection like `profiles[0]` — the list can change order, and `profile` forces replacement. A `lifecycle.precondition` fails at plan time if the pinned profile is no longer available, and `prevent_destroy` guards against accidental replacement:

```hcl
resource "clickhouse_service" "byoc" {
  profile               = "v1-standard-byoc-4"
  min_replica_memory_gb = 8 # must equal the profile's memory_gi
  max_replica_memory_gb = 8
  # ...

  lifecycle {
    precondition {
      condition     = contains([for p in data.clickhouse_service_profiles.byoc.profiles : p.profile], "v1-standard-byoc-4")
      error_message = "Profile v1-standard-byoc-4 is not available for this BYOC infra/region."
    }
    prevent_destroy = true
  }
}
```

Implementation: new `ListServiceProfiles` API client method; client mock regenerated with `make mock` (the bulk of the generated-line count in this PR).

## Testing / gates

- Unit tests: `syncServiceState` profile case, create-wiring tests (configured profile sent + synced; unknown profile omitted and settles to null), API client tests (path, `region_id`/`byoc_id` query params, empty `byoc_id` omitted), data-source mapping tests, and the registry golden data-source count bumped 12 → 13 per its own instructions.
- `make fmt`, `make lint`, `make test`, `make docs-check`, `make cover` all green.
- Docs regenerated; two-step example under `examples/data-sources/clickhouse_service_profiles/`.

Linear: BYOC-1138


nunoadrego (approved):

internal/api/service_profile.go
```diff
@@ -0,0 +1,47 @@
...
+
+// ListServiceProfiles returns the custom instance profiles available to the
+// organization in the given region. BYOC profiles are only included when
+// byocId is set to the BYOC infrastructure id they are configured for; in that
+// case regionId may be empty and the API derives the region from the
+// infrastructure.
+func (c *ClientImpl) ListServiceProfiles(ctx context.Context, regionId string, byocId string) ([]ServiceProfile, error) {
+	req, err := http.NewRequest(http.MethodGet, c.getOrgPath("/serviceProfiles"), nil)
+	if err != nil {
+		return nil, err
```
balanza: **suggestion:** ‏Consider error wrapping for better troubleshooting in case of error

internal/api/service_profile.go
```diff
@@ -0,0 +1,47 @@
...
+		q.Set("region_id", regionId)
+	}
+	if byocId != "" {
+		q.Set("byoc_id", byocId)
+	}
+	req.URL.RawQuery = q.Encode()
+
+	body, err := c.doRequest(ctx, req)
+	if err != nil {
+		return nil, err
```
balanza: **suggestion:** ‏error wrapping as above

internal/service/clickhouse/datasource/service_profiles.go
```diff
@@ -0,0 +1,157 @@
...
+		obj, d := types.ObjectValue(serviceProfileObjectType().AttrTypes, map[string]attr.Value{
+			"profile":   types.StringValue(it.Profile),
+			"cpu_cores": types.Float64Value(it.CpuCores),
+			"memory_gi": types.Float64Value(it.MemoryGi),
+		})
+		diags.Append(d...)
+		elems = append(elems, obj)
+	}
+	list, d := types.ListValue(serviceProfileObjectType(), elems)
+	diags.Append(d...)
```
balanza, line 65: **suggestion:** ‏We can avoid variable reassignment

```suggestion
	for _, it := range items {
		obj := types.ObjectValue(serviceProfileObjectType().AttrTypes
		diags.Append(map[string]attr.Value{
			"profile":   types.StringValue(it.Profile),
			"cpu_cores": types.Float64Value(it.CpuCores),
			"memory_gi": types.Float64Value(it.MemoryGi),
		}))
		elems = append(elems, obj)
	}
	list, d := types.ListValue(serviceProfileObjectType(), elems)
	diags.Append(d...)
```

balanza (approved):
Left some minor suggestions

internal/api/service_profile.go
luisliuchao: Fixed in `998065c` — wrapped as `failed to build list service profiles request: %w`.

internal/api/service_profile.go
luisliuchao: Fixed in `998065c` — wrapped as `failed to list service profiles: %w`.

internal/service/clickhouse/datasource/service_profiles.go
luisliuchao: Keeping the current form: `types.ObjectValue` returns `(ObjectValue, diag.Diagnostics)`, so the object and its diags can't be inlined into a single `diags.Append(...)` call without losing `obj` (the suggested patch doesn't compile as written). Each `d` here is a fresh short-declared variable per scope rather than a reassignment, and this mirrors the existing pattern in `datasource/service.go`. Happy to rename to `objDiags`/`listDiags` if you find that clearer.

balanza (approved):
Love it, thanks!

rndD (approved):

PR #704, opened 2026-09-12
feat(clickpipe): add initial_load_parallelism to MongoDB CDC settings

pfcoperez:
## Why

Expose the MongoDB initial-load parallelism setting added to the OpenAPI in ClickHouse/control-plane#30805 (`source.mongodb.settings.initialLoadParallelism`), so Terraform users can size the snapshot phase like the UI already allows. (DBI-696)

Closes: https://linear.app/clickhouse/issue/DBI-696

## What

- Add `source.mongodb.settings.initial_load_parallelism` (`Int64`, `>= 1`) to the `clickhouse_clickpipe` resource schema, model, API model, create request and read/state sync.
- The OpenAPI only accepts the value at creation time (it is excluded from `PATCH`), so the attribute is `Optional` + `Computed` with `RequiresReplace` and `UseStateForUnknown`, mirroring `snapshot_num_rows_per_partition` and `snapshot_number_of_parallel_tables`. When omitted, the API default (currently 4 for CDC initial load) is read back into state without producing a diff.
- Extend the MongoDB state-sync unit test to cover the value being read back and preserved as null when absent, update the MongoDB CDC example, regenerate docs.

## QA

- `go build ./...`, `go vet`, `go test ./internal/...` pass.
- `make docs` regenerated with Terraform 1.11.4 on PATH (the version CI uses); only the new attribute line changed.
- `make lint` reports one pre-existing gosec finding in `internal/api/common.go`, untouched by this change.
- The OpenAPI behaviour this relies on was validated end to end against the local playground in https://github.com/ClickHouse/control-plane/pull/30805#issuecomment-5646390220: explicit values and the default both land in the PeerDB catalog as `snapshot_max_parallel_workers`.
- Live acceptance against a deployed OpenAPI is pending the control-plane PR rollout.

Manual verification once the API is deployed:

1. Apply a MongoDB CDC pipe with `initial_load_parallelism = 3`; the state and `GET` echo 3.
2. Apply the same config without the attribute; state shows the API default (4) and a re-plan is empty.
3. Change the value on an existing pipe; the plan forces replacement.

### DIFF
--- a/internal/service/clickhouse/resource/models/clickpipe_resource.go
+++ b/internal/service/clickhouse/resource/models/clickpipe_resource.go
@@ -803,6 +803,7 @@ type ClickPipeMongoDBSettingsModel struct {
 	SyncIntervalSeconds            types.Int64  `tfsdk:"sync_interval_seconds"`
 	PullBatchSize                  types.Int64  `tfsdk:"pull_batch_size"`
 	ReplicationMode                types.String `tfsdk:"replication_mode"`
+	InitialLoadParallelism         types.Int64  `tfsdk:"initial_load_parallelism"`
 	SnapshotNumRowsPerPartition    types.Int64  `tfsdk:"snapshot_num_rows_per_partition"`
 	SnapshotNumberOfParallelTables types.Int64  `tfsdk:"snapshot_number_of_parallel_tables"`
 	DeleteOnMerge                  types.Bool   `tfsdk:"delete_on_merge"`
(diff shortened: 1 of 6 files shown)


internal/service/clickhouse/resource/models/clickpipe_resource.go
```diff
@@ -803,6 +803,7 @@ type ClickPipeMongoDBSettingsModel struct {
 	SyncIntervalSeconds            types.Int64  `tfsdk:"sync_interval_seconds"`
 	PullBatchSize                  types.Int64  `tfsdk:"pull_batch_size"`
 	ReplicationMode                types.String `tfsdk:"replication_mode"`
+	InitialLoadParallelism         types.Int64  `tfsdk:"initial_load_parallelism"`
```
pfcoperez, line 806: Selected Int64 for alignment with similar fields in other ClickPipes DBI connectors.

dtunikov (approved):

pfcoperez:
Merging as openapi has the required API changes in deployed in prod with the running version (https://github.com/ClickHouse/control-plane/commits/release/control-plane-api/260917.669?author=jgao54):

https://github.com/ClickHouse/control-plane/commit/3dbe7906090203d660f8167ab6b081d9a2db7cb2

